PR #12, opened 2016-02-11
Removed superfluous ref.

realzhtw:
I believe it's a simple typo.

### DIFF
--- a/include/gsl/gsl-lite.h
+++ b/include/gsl/gsl-lite.h
@@ -505,7 +505,7 @@ public:
         Expects( begin <= end );
     }
 
-    gsl_constexpr14 span( pointer & data, size_type size )
+    gsl_constexpr14 span( pointer data, size_type size )
         : begin_( data )
         , end_  ( data + size )
     {


martinmoene:
Thanks; 

However the `&` is needed for array size computation. 

(I undid the merge.)

martinmoene:
Updated previous comment.

realzhtw:
The problem here is that if it takes "pointer &" then you can't pass a temporary. Example:

    void f(span<int> s) { ... }

    ...
    int x;
    f(&x, 1);

So, maybe "const pointer &" then?

martinmoene:
I'm not sure temporaries should be allowed to feed a `span<>`.

Probably the constructor variant "C-array with size", `span( U (&arr)[N], size_type size )` has to go, so we can have `span( pointer data, size_type size )`.

I think it's best to wait for a/the specification of `span` from the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslview-views) to determine how to proceed.

realzhtw:
The  doc doesn't say much about it. But here is the excerpt from (https://github.com/Microsoft/GSL/blob/master/include/span.h)

    // construct from pointer + length
    constexpr span(pointer ptr, size_type size) noexcept : span(ptr, bounds_type{size}) {}

I can't see why construction from the temporary pointer is wrong. It the pointer that is temporary, not the data. So for example, construction like the following should be allowed (it's a temporary pointer):

    UglyVector<int> m(100);
    span<int> s(&m[0], 100);

Btw, there is also another useful constructor (which I in fact originally needed):

    constexpr span(reference data) noexcept : span(&data, bounds_type{1})

But I guess, you didn't added it because it's impossible to forbid construction from a temporary (data, not pointer) in C++0x, like it's done it Microsoft GSL:

    // prevent constructing from temporaries for single-elements
    constexpr span(value_type&&) = delete;

But thanks anyway. We're using your version of GSL in Tomtom, because we're still bound to C++03 in some places.

martinmoene:
>I can't see why construction from the temporary pointer is wrong. It is the pointer that is temporary, not the data. 

Agreed.

I based gsl-lite on Microsoft's GSL, however I'v not kept up with all their changes. I'm really eager to see the specification from [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslview-views) before changing too much.

Would it help you if I remove `span( U (&arr)[N], size_type size )` and change `span( pointer & data, size_type size )` to `span( pointer data, size_type size )`, i.e. remove the `&` ?

realzhtw:
> Would it help you if I remove span( U (&arr)[N], size_type size ) and change span( pointer & data, size_type size ) to span( pointer data, size_type size ), i.e. remove the & ?

Yes, this would help. Thanks in advance.
(Actually, we've imported the source to our repo and free to patch it, but it's always good to be as little away from upstream as possible.)

martinmoene:
Released [v0.2.0](https://github.com/martinmoene/gsl-lite/releases/tag/v0.2.0).

Thanks for your notes.

PR #16, opened 2016-03-10
Don't try to access cont[0] when cont is empty.

realzhtw:
gsl::span constructor should not try to access cont[0] when cont is empty when compiled with C++03.

### DIFF
--- a/include/gsl/gsl-lite.h
+++ b/include/gsl/gsl-lite.h
@@ -542,8 +542,8 @@ public:
         : begin_( cont.data() )
         , end_  ( cont.data() + cont.size() )
 #else
-        : begin_( &cont[0] )
-        , end_  ( &cont[0] + cont.size() )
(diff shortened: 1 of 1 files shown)


martinmoene:
Thanks @realzhtw. Fixed via 72d778e9ff78fdf16ab1f1c0992022a3d4cd5042

realzhtw:
Hi. I'm not sure if using ".empty()" instead of ".size() == 0" is OK. I couldn't find any documentation on what concept gsl::span is supposed to be able to construct from, but originally Microsoft implementation doesn't use .empty(), so requiring .empty() is an additional constraint.

martinmoene:
Thanks, I see your point and updated the code accordingly in dd106a06ac08f13ca7afd979b543e67cf04fa8f7. Indeed, we're waiting for the GSL _specification_.

PR #17, opened 2016-04-09
Fix compile error with clang targeting MSVC.

ned14:
You need a SFINAE disable for the implicit span<T> conversion from any type, else it is too greedy.


martinmoene:
Thanks!

PR #55, opened 2016-11-09
Free-standing functions: as_bytes, as_writable_bytes

ubique:
Microsoft Guideline Support Library has free-standing variant of as_bytes and as_writable_bytes. I have missed them switching from Microsoft GSL to gsl-lite(because of C++ standard downgrade).


martinmoene:
Thanks Dmitry.
[wow downgrade; I thought of gsl lite as a way forward]

Issue #59, opened 2016-11-10
Cleanup warnings


martinmoene:
See M-GSL [Cleanup additional warnings, PR 407](https://github.com/Microsoft/GSL/pull/407).

- [x] -Wcast-align
- [x] -Wconversion
- [x] -Wctor-dtor-privacy
- [x] -Wextra
- [x] -Wnon-virtual-dtor
- [x] -Wold-style-cast
- [x] -Woverloaded-virtual
- [x] -Wpedantic
- [x] -Wshadow (many in *lest*) => created [Issue at lest](https://github.com/martinmoene/lest/issues/59)
- [x] -Wsign-conversion, see PR #71

PR #71, opened 2017-04-23
Fix conversion and sign-conversion warnings.

Sillamacka:
Only two of the fixes were for GSL itself, but since all conversion warnings
have been fixed in the tests and these warnings have been enabled for the test
suite, it should be easier to discover conversion warnings in GSL in the future.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -59,7 +59,7 @@ elseif( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" OR
         target_compile_options( gsl-lite-cpp14.t  PUBLIC -std=c++14 )
     endif()
 
-    add_compile_options( -Wall -Wno-missing-braces -fno-elide-constructors )
+	set ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-missing-braces -fno-elide-constructors -Wconversion -Wsign-conversion -Wno-string-conversion" )
 
 elseif( ${CMAKE_CXX_COMPILER_ID} MATCHES "Intel" )
 # as is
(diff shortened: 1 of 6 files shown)


martinmoene:
Thanks @Sillamacka 

(ref. issue #59)

test/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ elseif( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" OR
         target_compile_options( gsl-lite-cpp14.t  PUBLIC -std=c++14 )
     endif()
 
-    add_compile_options( -Wall -Wno-missing-braces -fno-elide-constructors )
+	set ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-missing-braces -fno-elide-constructors -Wconversion -Wsign-conversion -Wno-string-conversion" )
```
martinmoene: What is the reason to change `add_compile_options` to `set` ?

Sillamacka: I'll be honest, I don't fully understand the difference, but when using `add_compile_options` no warnings were thrown and `make VERBOSE=1` did not show the warning settings on the command line.

Sillamacka: I looked it up, this is the issue with `add_compile_options`:

> Adds options to the compiler command line for targets in the current directory and below that are added **after** this command is invoked.

[From this stack overflow answer](http://stackoverflow.com/a/40522384)
Isn't CMake wonderfully intuitive?

So it seems the `CMakeLists.txt` needs to be a bit rearranged in general to fix this.

martinmoene: Aha, thanks!

PR #72, opened 2017-04-26
Support for output stream printing of string_span types via operator<<

Sillamacka:
Following the behaviour described for `std::basic_string` [here](http://en.cppreference.com/w/cpp/string/basic_string/operator_ltltgtgt).

Unfortunately, I had to remove the specialised `string_span` printing implementation from the test suite as it would otherwise conflict.


martinmoene:
Note to self: [GSL issue 160](https://github.com/Microsoft/GSL/issues/160)

PR #92, opened 2017-12-11
Make v0.26.0 available in conan

agauniyal:
https://bintray.com/agauniyal/nonstd-lite/gsl-lite%3Anonstd-lite/0.26.0%3Astable

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Or, if you use the [conan package manager](https://www.conan.io/), follow these 
 2. Add a reference to *gsl-lite* to the *requires* section of your project's `conanfile.txt` file:
 
         [requires]
-        gsl-lite/0.25.0@nonstd-lite/stable
+        gsl-lite/0.26.0@nonstd-lite/stable
 
 3. Run conan's install command:
 


agauniyal:
Btw here's how I'm testing my libraries over different compilers and their versions - https://github.com/agauniyal/termdb/blob/master/.travis.yml#L1-L104 .You can copy paste the entire highlighted block and rest of your scripts section will work fine with it :)

martinmoene:
Thanks, will have a look at it.

agauniyal:
@martinmoene would it be fine to request a string_view backport 😬 ? Many other OSS implementations provide variant and optional backports to c++11 and apart from them I really want to recommend using string_view to people(works wonders in function parameters 👌 ) but I can't find a single decent implementation online 😞

martinmoene:
@agauniyal one can always ask ;)

I [found on GitHub](https://github.com/search?l=C%2B%2B&q=string+view&type=Repositories&utf8=%E2%9C%93):
- @bitwizeshift, Matthew Rodusek's [string_view Standalone](https://github.com/bitwizeshift/string_view-standalone)
- @satoren 's [string_view for C++03 C++11 C++14](https://github.com/satoren/string_view)

I may give it a try...

martinmoene:
..it may take a short while to start on [string-view-lite](https://github.com/martinmoene/string-view-lite).

martinmoene:
@agauniyal I'm using your [travis configuration suggestion above](https://github.com/martinmoene/gsl-lite/pull/92#issuecomment-350710030) for [string-view-lite](https://github.com/martinmoene/string-view-lite). However it seems that the same compiler version is used for each of gcc and clang. The same holds for [termdb](https://travis-ci.org/agauniyal/termdb).

Do you have an idea / can you explain what's happening?

agauniyal:
@martinmoene I think cmake is using `CXX` over `COMPILER` env variable, setting `export CXX=$COMPILER` should make it work again.

At line [794](https://travis-ci.org/martinmoene/string-view-lite/jobs/320777886#L794) `COMPILER` is set to right version of `clang`, but then at line [798](https://travis-ci.org/martinmoene/string-view-lite/jobs/320777886#L798) travis sets `CXX` which gets used by cmake. So if you set `CXX` again to `COMPILER` right before cmake is called it should use right version. In my case I do it here - https://github.com/agauniyal/termdb/blob/master/.travis.yml#L120

martinmoene:
@agauniyal Thanks, that helps! I now also see in the runs of termdb which compiler is used.

Issue #38, opened 2016-09-21
Adapt CmakeLists.txt for easier consumption of GSL Lite by another project


martinmoene:
Changes:
- Use `target_include_directories` over `include_directories`.
- Use `INTERFACE` library.
- Add GSL project to `tests/CMakeLists.txt`.

**Note**: requires CMake 2.8.12 instead of 2.8.11.2 (Travis).

See M-GSL [PR #370](https://github.com/Microsoft/GSL/pull/370).

nholthaus:
In reality, using interface libraries may require CMake 3.3, at least if your library clients want to be able to depend on them, due to [this bug](https://cmake.org/Bug/view.php?id=15414). This [SO question](http://stackoverflow.com/questions/35630755/propagate-dependencies-to-header-only-externalproject-with-cmake) may also be informative.

PR #100, opened 2018-02-18
CMake package configuration

FlorianWolters:
I've added [CMake Package](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html) support for gsl-lite, which makes it easy to install and consume gsl-lite as a package with the build system CMake.

### Installation

Hint: I am using MSVS 2017 and Ninja

    cmake -H. -B../_build -G"Ninja" -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=cl -DCMAKE_INSTALL_PREFIX="D:/dev/native/gsl-lite/gsl-lite-0.29.0"
    cmake --build ../_build --config Release --target install

### Consumption

Consider a package `package_using_gsl-lite` that would like to use v0.29.0 of the package `gsl-lite`.

`CMakeLists.txt`:

    cmake_minimum_required(VERSION 3.0 FATAL_ERROR)

    find_package(gsl-lite 0.29 REQUIRED)

    project("package_using_gsl-lite" LANGUAGES CXX)

    add_executable("package_using_gsl-lite" main.cpp)
    target_link_libraries("package_using_gsl-lite"
      PRIVATE gsl-lite::gsl-lite
      )

`main.cpp`:

    // Standard Library
    #include <iostream>
    #include <string>

    // Guideline Support Library (GSL)
    #include <gsl/gsl>

    void not_null_example(const gsl::not_null<const std::string*> arg) {
      // NOOP
    }

    int main() {
      using UniqueImmutableStringPtr = std::unique_ptr<const std::string>;
      const UniqueImmutableStringPtr text = std::make_unique<std::string>("hello, gsl");
      not_null_example(text.get());
      //not_null_example(nullptr);  

      std::cout << text.get() << '\n';
    }

The variable `gsl-lite_DIR` can be used to specify the path to the CMake Package Configuration file of gsl-lite.

    cmake -H. -B../_build -G"Ninja" -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=cl -DCMAKE_INSTALL_PREFIX=_stage -Dgsl-lite_DIR="D:/dev/native/gsl-lite/gsl-lite-0.29.0/lib/cmake/gsl-lite"
    cmake --build ../_build --config Release

This resolves issue #38, imo. I've implemented it in a modern CMake-idiomatic and simple way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,20 +5,66 @@
 #
 # This code is licensed under the MIT License (MIT).
 
-cmake_minimum_required( VERSION 3.0 )
+cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 
-project( gsl_lite )
+include(CMakePackageConfigHelpers)
 
-include( GNUInstallDirs )
+set(package_name "gsl-lite")
 
-add_library(    gsl_lite INTERFACE )
-target_sources( gsl_lite INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/gsl-lite.natvis )
+project(${package_name} VERSION 0.29.0 LANGUAGES CXX)
 
+set(include_source_dir "${PROJECT_SOURCE_DIR}/include")
+set(include_install_dir "include")
+set(package_config_install_dir "lib/cmake/${PROJECT_NAME}")
+set(package_config_file "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake")
+set(package_version_file
+  "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
+  )
+set(package_export_name "${PROJECT_NAME}Targets")
+set(package_namespace "${PROJECT_NAME}::")
+
+# Interface library
+add_library(${package_name} INTERFACE)
+target_include_directories(${package_name}
+  INTERFACE
+    "$<BUILD_INTERFACE:${include_source_dir}>"
+  )
+target_sources(${package_name}
+  INTERFACE "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/gsl-lite.natvis>"
+  )
+
+# Package configuration
+configure_package_config_file(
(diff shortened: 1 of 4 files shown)


martinmoene:
Thanks for your contribution Florian.

I'm not (yet) familiar with CMke packages.

What are your expectations now, given CI breaks?

FlorianWolters:
Ugh, it would have been better if I had completely ignored the `CMakeLists.txt` in the `test` directory instead, because that file is independent from the CMake package configuration.

Previously the `test/CMakeLists.txt` file seemed to rely that `$ENV{VS2017INSTALLDIR}/VC/Auxiliary/VS/include` is in the include path (see `test/gsl-lite.t.hpp`) There was no `find_package` or `find_path` command trying to locate that MSVC specific include path. I've tried to explicitly find the directory path to the `CppCoreCheck/warnings.h` file, which works on my machine, but fails on CI.

My solution is buggy:
1. The environment variable `VS2017INSTALLDIR` does not seem to exist on AppVeyor.
2. The `find_path` is always executed, regardless of the compiler used. The command should be executed if using `MSVC >= 1910` only.

Both the old approach and my approach have the same problem. It isn't the "CMake"-way to rely on the environment. Instead a CMake option should be used during the configure phase. Using an explicit `-DCppCoreCheck_DIR="<path>"` would be better, though I do not know how-to obtain the directory path to the MSVC-CppCoreCheck directory on AppVeyor.

Maybe I should submit another PR which adds package configuration only. For that it is sufficient to modify the top-level `CMakeLists.txt` file.

Alternatively you can make the changes in the top-level `CMakeLists.txt` by yourself, simply using the file submitted in this PR.

martinmoene:
@FlorianWolters  ok, thanks, I'll look into it further.

martinmoene:
ToDo:
- [ ] Only build and perform tests if toplevel CMakeFiles.txt.
- [ ] Add example (.cpp, CMakeFiles.txt) that uses the package (see [above](https://github.com/martinmoene/gsl-lite/pull/100#issue-169854818))
- [ ] Update Readme with relevant info

PR #107, opened 2018-03-07
Introduce support for DJGPP cross compiler

zemasoft:
This compiler is GCC based and is targeting DOS. It does not support `wchar_t` nor `std::wstring` probably because there would not be much use of them in DOS.

All the tests and examples were tested on DOSBox and on FreeDOS using DJGPP for GCC 7.2.0 with success (actually there is one test failing in *cpp11 and in *cpp14 but you are probably aware of that).

Since DJGPP is a cross compiler, user has to supply cmake with `-DCMAKE_SYSTEM_NAME=Generic` to be able to build the project with it, see https://cmake.org/Wiki/CMake_Cross_Compiling.

Some relevant links regarding DJGPP cross compiler:
https://www.youtube.com/watch?v=yHrC_rZUaiA
https://github.com/andrewwutw/build-djgpp

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -139,6 +139,14 @@
 # define gsl_COMPILER_GNUC_VERSION  0
 #endif
 
+// Presence of wide character support:
+
+#ifdef __DJGPP__
+# define gsl_HAVE_WCHAR 0
+#else
+# define gsl_HAVE_WCHAR 1
+#endif
+
 // Presence of C++11 language features:
 
 #define gsl_CPP11_10  (gsl_CPP11_OR_GREATER || gsl_COMPILER_MSVC_VERSION >= 10)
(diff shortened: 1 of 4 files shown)


martinmoene:
Hi Tomáš,

>(actually there is one test failing in *cpp11 and in *cpp14 but you are probably aware of that).

Can you tell which one test is failing? 

On Travis all tests pass with all compilers that support -std=c++11/14. _Examples_ n the other hand may fail for educational reasons.

martinmoene:
@zemasoft , do you perhaps mean:
```
finally: Allows moving final_action to throw[.]
```

zemasoft:
Hi Martin,

Yes, this is what I see:
```
gsl-lite/test/util.t.cpp:136: failed: finally: Allows moving final_action to throw[.]: i == 1 for 0 == 1
1 out of 281 selected tests failed.
```
Am I misreading it somehow? Or is this intentional? But the error level returned by the test is `1` so...

zemasoft:
But when looking into `test/CMakeLists.txt`, I should probably not run the tests with `"*"` but just without any argument. Am I right?

martinmoene:
Yes, quite right!

Normal test run is just e.g.:
> prompt\> gsl-lite-cpp11.t

However your remark that there is one silenced (!) test that fails still stands to be sure. I've not yet worked out what to do about it (and lost sight that it's to be done).

zemasoft:
And is it necessary for the `struct action` to have the copy constructor unavailable/deleted? Because if not, `std::is_nothrow_move_constructible` might then be used to adapt `finally()` and `class final_action`.

But take it just as my naive opinion...

PR #126, opened 2018-05-22
Fix final_action_return and final_action_error during stack unwinding

patstew:
Previously they would always take the error path, even when there is no exception in the guarded function. This is the reason why std::uncaught_exception was deprecated in favour of uncaught_exceptions, there's a paper about it here: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4152.pdf
I've added uncaught_exceptions for pre-2017 MSVC, GCC and Clang which is basically the same as the one used here: https://github.com/evgeny-panasyuk/stack_unwinding/blob/master/boost/exception/uncaught_exception_count.hpp
That version has been tested with a number of compiler versions.

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -681,18 +681,6 @@ protected:
         invoke_ = false;
     }
 
-#if gsl_CPP17_OR_GREATER
-    gsl_api int uncaught_exceptions() gsl_noexcept
-    {
-        return std::uncaught_exceptions();
-    }
-#else
-    gsl_api int uncaught_exceptions()
-    {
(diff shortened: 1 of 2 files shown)


martinmoene:
Thanks @patstew .
___

Notes to self:

Post-merge:
- [ ] Move `uncaught_exceptions()` to its own `#if gsl_FEATURE(EXPERIMENTAL_RETURN_GUARD)` to be accessible further down.
- [ ] Express as `#if gsl_HAVE(UNCAUGHT_EXCEPTIONS) ... #elif .... #elif ....#endif`.
- [ ] Reformat code.

PR #149, opened 2019-03-22
Update conan package documentation

nathanaelg:
These two changes should probably be in different PRs but they are small changes that you may be happy to bring in together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,12 +221,12 @@ For the [conan package manager](https://www.conan.io/), follow these steps:
 
 1. Add *nonstd-lite* to the conan remotes:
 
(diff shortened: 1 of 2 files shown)


martinmoene:
@nathanaelg Thanks for the PR.

How would you explain the change from

>`self.copy("*.hpp")`

to:

>`self.copy("include/gsl/gsl")`

nathanaelg:
It's an addition rather than a change, the `"*.hpp"` matches `include/gsl.hpp`, `include/gsl/gsl-lite-vc6.hpp` and `include/gsl/gsl-lite.hpp` but doesn't match the `include/gsl/gsl` file which gets used for compatibility with MS GSL.

martinmoene:
Ah thanks, I see.

PR #152, opened 2019-04-06
Add support for user-defined contract violation handler

petamas:
This is useful if user already has a custom assert handling system. (My current company has one, and integrating GSL with it can help company-wide adoption.)

Inspired by Boost.Assert's similar mechanism: https://www.boost.org/doc/libs/1_69_0/libs/assert/doc/html/assert.html

Decided to exclude gsl::narrow() because both gsl_CONFIG_CONTRACT_LEVEL_OFF and GSL_UNENFORCED_ON_CONTRACT_VIOLATION excludes it, so it seemed consistent this way.


martinmoene:
Add via 973328b0a8ee8e1c7cffd74903b9f139ea3a9e93, b7eefba5614e5b14561d5b2b97776b4e636568a9 and a548a7fe7d9afee5204e5b6ba6be8e2a818a467d

mbeutel:
Addressed by https://github.com/martinmoene/gsl-lite/commit/a548a7fe7d9afee5204e5b6ba6be8e2a818a467d.

PR #158, opened 2019-05-13
add support for std::initializer_list

### LINKED ISSUE #157: Cannot construct span from initializer_list
Hello,

It is not currently possible to construct a `gsl::span` from a `std::initializer_list`, with the latest master version.

It is because `std::initializer_list` does not have a `.data()` member function.

It does work with the latest clang and `std::span`: https://godbolt.org/z/DtTwal

I believe reimplementing `std::data` and `std::size` for C++14 and lower should do the trick.

theodelrieu:
We must emulate `std::data/std::size`, since `initializer_list` does
not have a `data()` member function.

Additionally, we must pass `Container&/const Container&` to
`detail::can_construct_span_from`, so that `declval` returns the correct
reference type.

Fixes #157

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -418,6 +418,11 @@
 # include <cstdint>
 #endif
 
+#if gsl_HAVE( INITIALIZER_LIST )
+# include <initializer_list>
+#endif
+
+
 // MSVC warning suppression macros:
 
 #if gsl_COMPILER_MSVC_VERSION >= 140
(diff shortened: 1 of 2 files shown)


martinmoene:
@theodelrieu  I've added construction from `std::initializer_list` to project [nonstd::span](https://github.com/martinmoene/span-lite); will look to move over the changes there to *gsl lite* here later.

theodelrieu:
Oh, I did not know this project existed... Will take a look, thanks!

PR #165, opened 2019-07-13
Suppress more warnings I get from my project from this include.

AraHaan:
This should help reduce my warnings to that of only what is mine / stdlib ones when I use modules TS (that part I seem to not be able to control).


AraHaan:
Also warns with ``gsl-lite\include\gsl\gsl-lite.hpp(1167): C6385: Reading invalid data from 'arr':  the readable size is '255' bytes, but 'pos' bytes may be read.``

(The number is based on the top of these modifications) To test it; first apply this.

PR #167, opened 2019-08-04
cmakelint CMakeList.txt files

ClausKlein:
prevent warnings with apple clang++ V8.0.0
prevent core dump at example

### DIFF
--- a/test/gsl-lite.t.hpp
+++ b/test/gsl-lite.t.hpp
@@ -31,7 +31,9 @@
 #ifdef __clang__
 # pragma clang diagnostic ignored "-Wstring-conversion"
 # pragma clang diagnostic ignored "-Wunused-parameter"
+#ifndef __APPLE_CC__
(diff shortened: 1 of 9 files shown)


martinmoene:
Thanks for your contribution. It will take a few weeks before a will be able to process it as I'm currently on holydays.

test/gsl-lite.t.hpp
```diff
@@ -31,7 +31,9 @@
 #ifdef __clang__
 # pragma clang diagnostic ignored "-Wstring-conversion"
 # pragma clang diagnostic ignored "-Wunused-parameter"
+#ifndef __APPLE_CC__
```
mbeutel: What exactly is the issue here? Does Apple's flavor of Clang not support `-Wunused-template`?

mbeutel:
I'm closing this now because it is based on a rather old revision. Some of the formatting changes have already found their way into other PRs. I still welcome any comments on my review.

PR #174, opened 2019-08-26
Remove function specifiers for defaulted functions

codecircuit:
Up to now, only `__host__` `__device__` specifiers were
added with the gsl_api macro. `nvcc` also warns about
these function specifiers.

See also the discussion here:
https://stackoverflow.com/questions/46469062/warning-host-annotation-on-a-defaulted-function-is-ignored-why

Maybe you prefer to create an additional macro for defaulted functions.


martinmoene:
Let's wait with additional macro for defaulted functions until the need arises.

Issue #161, opened 2019-06-10
Barely usable definition of fail_fast_assert_handler


stohrendorf:
`fail_fast_assert_handler` is declared with `gsl_constexpr14`, which doesn't make it really useful as you're required to only have constexpr statements in its body; also, it requires the function definition to be inline. I'm not sure if the handler actually needs to be `constexpr`, but I'd like to implement a handler like this, which is currently not possible (note the `gsl_constexpr14` to have a compatible signature, which doesn't work together with `BOOST_LOG_TRIVIAL`, also I'd like to have this function instantiated only in a single translation unit as performance is not important):

```
gsl_api gsl_constexpr14 void
gsl::fail_fast_assert_handler(char const* const expression, char const* const message, char const* const file, int line)
{
    BOOST_LOG_TRIVIAL( error ) << "Expectation failed at " << file << ":" << line;
    BOOST_LOG_TRIVIAL( error ) << "  - expression " << expression;
    BOOST_LOG_TRIVIAL( error ) << "  - message " << message;
    BOOST_LOG_TRIVIAL( error ) << "Stacktrace:\n" << boost::stacktrace::stacktrace();
    BOOST_THROW_EXCEPTION( gsl::fail_fast(message) );
}
```

stohrendorf:
Thank you very much.

PR #176, opened 2019-09-12
Mixed bag of minor improvements

### LINKED ISSUE #133: use iosfwd instead of ostream
I tested replacing the `<ostream>` with `<iosfwd>` and found no issue.
Given that the templates only forward calls there should be no reason to pull in `<ostream>`, which is rather huge. 

Should a template be instantiated then its required that the called method pulls in `<ostream>` (anything else would be a violation of the standard and probably wont compile)

mbeutel:
- use `constexpr` in more places
- support exporting build directory to user package registry
- fix #133

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ endif()
 
 option( GSL_LITE_OPT_BUILD_TESTS    "Build and perform gsl-lite tests" ${gsl_IS_TOPLEVEL_PROJECT} )
 option( GSL_LITE_OPT_BUILD_EXAMPLES "Build gsl-lite examples" OFF )
+option( CMAKE_EXPORT_PACKAGE_REGISTRY "Export build directory to CMake user package registry" OFF )
 
 # If requested, build and perform tests, build examples:
 


petamas:
Hi @mbeutel and  @martinmoene,

this PR seems to directly contradict issue #161: in the resolution to that issue, constexpr qualifiers were removed from fail_fast_assert_handler and related functions, but now they're back. What to do with #161 then? (I'm mostly interested in this since I'm the original creator of fail_fast_assert_handler.)

mbeutel:
> this PR seems to directly contradict issue #161: in the resolution to that issue, constexpr qualifiers were removed from fail_fast_assert_handler and related functions, but now they're back. What to do with #161 then? (I'm mostly interested in this since I'm the original creator of fail_fast_assert_handler.)

Note that https://github.com/martinmoene/gsl-lite/pull/176/commits/74f5f76dc4adb42128f3c194aec2ad0bac46296d in #176 does *not* add `gsl_constexpr14` to `fail_fast_assert_handler()`, only to the various flavors of `fail_fast_assert()`. This makes it possible to use contract checks in `constexpr` functions. The ability to perform non-`constexpr` operations in `fail_fast_assert_handler()` is unimpaired.

Example:
https://gcc.godbolt.org/z/NmUzaj

petamas:
@mbeutel: Thanks for the quick reply! I didn't spot that difference, but now it's clear. Your example code was also a revelation, I did not know you can conditionally call non-constexpr functions from constexpr functions if the condition is false in constexpr contexts. I think I need to learn a bit more of C++14. :) Consider my issue dropped.

Issue #184, opened 2019-09-18
not_null's dereference operator does not work for smart pointers with c++11


petamas:
The second and third test case does not compile when gcc is set to C++11 mode:

```
using point = std::pair<int, int>;

TEST(GslNotNull, NNRawDereference)
{
	not_null<point*> nnp(new point(1, 3));
	EXPECT_EQ(*nnp, std::make_pair(1,3));
}

TEST(GslNotNull, NNUniqueDereference)
{
	not_null<unique_ptr<point>> nnp(new point(1, 3));
    EXPECT_EQ(*nnp, std::make_pair(1,3));
}

TEST(GslNotNull, NNSharedDereference)
{
	not_null<shared_ptr<point>> nnp(new point(1, 3));
    EXPECT_EQ(*nnp, std::make_pair(1,3));
}
```

See: https://gcc.godbolt.org/z/TtEth-

The problem is that operator* is defined using decltype(auto), which is available from C++14. It should be replaced with something like `decltype(*declval<T>())` or with something along the lines of Dropbox nn:
```
namespace nn_detail {
template <typename T> struct element_type { using type = typename T::element_type; };
template <typename Pointee> struct element_type<Pointee *> { using type = Pointee; };
}
// ...
using element_type = typename nn_detail::element_type<PtrType>::type;
// ...
element_type & operator*()  const { return *ptr; }
```

The declval approach is problematic, because it doesn't work with C++03; in contrast, the nn-style approach compiles with C++03, but requires smart pointers to have an embedded `element_type` typedef.

mbeutel:
Thank you for the analysis. Yes, I think we can do better than defining `operator *()` for C++14 only.

Perhaps we can use `PtrT::element_type` if it exists, and `decltype(*declval<T>())` otherwise (if C++11).

petamas:
I was thinking the reverse: `if(C++11) { use decltype } else { use PtrT::element_type }` - this way we only add restrictions to PtrT in pre-modern C++, where smart pointers are used much less. Also, checking whether PtrT::element_type exists can be quite tricky, so GSL's code would be easier to understand this way.

mbeutel:
We wouldn't be adding restrictions to `PtrT` in C++11 because we'd check if `PtrT::element_type` exists, and only use it if it does. By trying `PtrT::element_type` first, we keep the return type selection stable with respect to the C++ language version (though admittedly `PtrT::element_type` is unlikely to differ from `decltype(*declval<T>())`).

[...]

PR #192, opened 2019-09-23
Fix issue #184: not_null's dereference operator did not work for smart pointers with c++11

petamas:


include/gsl/gsl-lite.hpp
```diff
@@ -1293,7 +1296,38 @@ class not_null
     typedef T get_result_t;
 #endif
 
+#if gsl_CPP11_OR_GREATER
```
mbeutel: We define `void_t<>` only if `gsl_CPP11_120`, so perhaps this section should also query for `gsl_CPP11_120`.

include/gsl/gsl-lite.hpp
```diff
@@ -1293,7 +1296,38 @@ class not_null
...
+    template<class U, class E = void>
+    struct element_type_helper
```
mbeutel: Could you perhaps move `element_type_helper<>` out-of-class (preferably to namespace `gsl::detail`)? It doesn't depend directly on the template arguments of `not_null<>`, so keeping it outside potentially reduces the number of specializations the compiler has to generate. Also, I remember VS 2015 having occasional trouble with in-class specializations.

include/gsl/gsl-lite.hpp
petamas: gsl_CPP11_120 is _wider_ than gsl_CPP11_OR_GREATER, it includes every C++11-compatible compiler plus Visual Studio 2013. However, I do not know if VS2013 supports decltype and declval, so I think it's safer this way.

include/gsl/gsl-lite.hpp
petamas: ok

include/gsl/gsl-lite.hpp
mbeutel: Right, I missed that.
The tables at https://en.cppreference.com/w/cpp/compiler_support are useful to check compiler support for language/library features. It seems that `decltype` was added in VS2015, so `gsl_CPP11_OR_GREATER` is fine.

petamas: Thanks for the table, I did not know that page!

But I think you misread it: decltype(auto) was added in 19.0 (VS2015), but "simple" decltype was added in 16.0 (VS2010), a.k.a. gsl_CPP11_100, so theoretically, we could use gsl_CPP11_120. However, declval() is not in the table, and I do not have access to VS2010 to try it out, so I still think we should stick to gsl_CPP11_OR_GREATER.

mbeutel: Right again. Today isn't my day.
I don't think we should put too much energy into supporting C++11-related features on versions of Visual Studio older than VS2015.

mbeutel:
Merged, thank you.

martinmoene:
Remember section [_A.3 GSL Lite test specification_](https://github.com/martinmoene/gsl-lite#a3-gsl-lite-test-specification) of the Readme :)

I made the tests prominently visible here not in the least to spread Kevlin Henney's idea to use predicates for tests. See for example [Programming with GUTs by Kevlin Henney](https://www.youtube.com/watch?v=azoucC_fwzw).

petamas:
@martinmoene : Dangit, there was a point in time I thought about updating this, but it seems I forgot eventually. I'll fix in my next PR (which will add quite a number of new tests about unique_ptr and shared_ptr).

Is there a way to generate it automatically? I haven't really looked at the options of LEST yet.

martinmoene:
@petamas , I've always generated and updated the list by hand, and (also) often forgot to do so :)

On Windows, I did the equivalent of
```
cd ...\gsl-lite\test & t.bat & gsl-lite.t.exe -l | winclip -c
```
and pasted the clipboard at the right spot in `Readme.md`

`t.bat` compiles with `VC` and produces a list of tests in the desired order, whereas `tg.bat` compiles with `gcc` and produces a list of tests that is meh.

`winclip` is part of the [outwit tool set](https://www.spinellis.gr/sw/outwit/) by [Diomidis Spinellis](https://www.spinellis.gr/).

mbeutel:
Martin, thanks for bringing this up. I wondered about the list myself (and I also forgot to update it back when I added `ExpectsAudit()`/`EnsuresAudit()`).


> `t.bat` compiles with `VC` and produces a list of tests in the desired order, whereas `tg.bat` compiles with `gcc` and produces a list of tests that is meh.

I guess I'll update these sometime so they use CMake rather than invoking `cl`/`g++` directly. Oh, and add `t.sh` :)


> `winclip` is part of the [outwit tool set](https://www.spinellis.gr/sw/outwit/) by [Diomidis Spinellis](https://www.spinellis.gr/).

(BTW, Windows Vista and newer comes with a built-in clipboard tool called `clip`. Piping text to `clip` puts it in the clipboard.)

martinmoene:
> (BTW, Windows Vista and newer comes with a built-in clipboard tool called clip. Piping text to clip puts it in the clipboard.)

Ah, thanks!

petamas:
@martinmoene : Thanks, t.bat (and the other steps) worked like a charm. (After I ran it from a Developer Command Prompt - cl is not in my path variable.)

@mbeutel : CMake-ing it sounds nice. :)

Issue #129, opened 2018-05-30
Add test with not_null<unique_ptr<T>>


martinmoene:
See M-GSL PR 675.

petamas:
I've been playing around with not_null<unique_ptr<T>>, and found a few problems.

I'll use these helpers in my examples:
`
using point = std::pair<int, int>;

template <typename T>
using nn_unique_ptr = gsl::not_null<std::unique_ptr<T>>;

template <typename T, typename... Args>
nn_unique_ptr<T> nn_make_unique(Args &&... args) {
	return nn_unique_ptr<T>(std::unique_ptr<T>(new T(std::forward<Args>(args)...)));
}
`

### operator-> is not working by default
This code does not compile:
`
TEST(GslNotNull, NNMakeUniqueWorks)
{	
	nn_unique_ptr<point> nnp = nn_make_unique<point>(1, 3);
	EXPECT_EQ(nnp->first, 1);
	EXPECT_EQ(nnp->second, 3);
}
`
See: https://gcc.godbolt.org/z/GmDX_I

The problem is that operator-> tries to return a unique_ptr<point> by value, but since it is move-only, it cannot. If we use -Dgsl_CONFIG_NOT_NULL_GET_BY_CONST_REF, this problem goes away.

### not_null<unique_ptr<T>>  cannot be converted to not_null<unique_ptr<const T>>

This code does not compile (not even with -Dgsl_CONFIG_NOT_NULL_GET_BY_CONST_REF):
`
TEST(GslNotNull, NNUniqueCanBeConvertedToConstUnique)
{
	nn_unique_ptr<const point> nnp = nn_make_unique<point>(1, 3);
	EXPECT_EQ(nnp->first, 1);
	EXPECT_EQ(nnp->second, 3);
}
`

See: https://gcc.godbolt.org/z/nNXvuC

The problem is that the not_null<U> constructor has a const& parameter,  it doesn't have a "move" version. Similarly, get() lacks a "moving" version.

### not_null<unique_ptr<T>> cannot be converted to not_null<shared_ptr<T>>

This code does not compile (not even with -Dgsl_CONFIG_NOT_NULL_GET_BY_CONST_REF):
`
TEST(GslNotNull, NNUniqueCanBeConvertedToShared)
{
	nn_shared_ptr<point> nnp = nn_make_unique<point>(1, 3);
	EXPECT_EQ(nnp->first, 1);
	EXPECT_EQ(nnp->second, 3);
}
`

[...]

Issue #178, opened 2019-09-18
Cleanup for 1.0


mbeutel:
There are several occasions for code cleanup if we decide to make some breaking changes in a major version update.

- [x] Introduce a clear distinction between internal macros (implementation details, subject to change) and public macros (intended for public consumption), e.g. by giving internal macros a trailing `_`.
- [x] Remove currently unused macros: `gsl_REQUIRES_0()`, `gsl_REQUIRES_R()`, `gsl_DIMENSION_OF()`, probably more
- [x] Remove `gsl_REQUIRES_T()` in favor of `gsl_REQUIRES_A()`, and perhaps change that to the more idiomatic `typename enable_if<..., int>::type = 0`
- [x] Change default of `gsl_FEATURE_OWNER_MACRO` to 0 (only needed for C++98).
- [x] Change default of `gsl_CONFIG_ALLOWS_NONSTRICT_SPAN_COMPARISON` to 0 (consistent with C++20 spans).
- [x] `final_action<>` should not need a virtual destructor, and `final_action_return<>` and `final_action_error<>` should not inherit from `final_action<>`.
- [x] Change default of `gsl_CONFIG_NOT_NULL_EXPLICIT_CTOR` to 1 (thanks @petamas)

petamas:
I'd also consider changing the default of gsl_CONFIG_NOT_NULL_EXPLICIT_CTOR to 1 (or removing the config option completely) - there's a not_null_ic class if someone wants the implicit constructor, and making it explicit ensures more type-safety (users have to manually write not_null, making them declare in writing that they are sure it won't be null).

mbeutel:
Right, I missed that. Added it now, thanks.

PR #197, opened 2019-09-27
Fix issue #129: not_null<unique_ptr<T>> is now working as expected

petamas:
Also:
- unit tests from earlier PRs #176 and #192 are added to README (along with tests added in this one)
- fixed -Wunnamed-type-template-args warnings introduced in PR #192

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -296,6 +296,7 @@
 #define gsl_HAVE_AUTO                   gsl_CPP11_100
 #define gsl_HAVE_NULLPTR                gsl_CPP11_100
 #define gsl_HAVE_RVALUE_REFERENCE       gsl_CPP11_100
+#define gsl_HAVE_FUNCTION_REF_QUALIFIER ( gsl_HAVE_RVALUE_REFERENCE && ! gsl_BETWEEN( gsl_COMPILER_GNUC_VERSION, 1, 481 ) )
 
 #define gsl_HAVE_ENUM_CLASS             gsl_CPP11_110
 
@@ -1318,6 +1319,9 @@ gsl_api inline gsl_constexpr T & at( span<T> s, size_t pos )
 // not_null<> - Wrap any indirection and enforce non-null.
 //
 
+template< class T >
+class not_null;
+
 namespace detail
 {
     // helper class to figure out the pointed-to type of a pointer
@@ -1350,6 +1354,14 @@ namespace detail
     };
 #endif
 
+    template<class T>
+    struct is_not_null : public std11::false_type {};
+
+    template<class T>
+    struct is_not_null< not_null< T > > : public std11::true_type {};
+
+    template<class T>
+    struct is_not_null_cvref : public is_not_null< typename std20::remove_cvref< T >::type > {};
 }
 
 template< class T >
@@ -1367,6 +1379,9 @@ class not_null
     typedef T get_result_t;
 #endif
 
+    template< class U >
+    friend class not_null;
+
 public:
     typedef typename detail::element_type_helper<T>::type element_type;
 
@@ -1374,14 +1389,21 @@ public:
     static_assert( std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr." );
 #endif
 
-    template< class U
-#if gsl_HAVE( DEFAULT_FUNCTION_TEMPLATE_ARG )
-        gsl_REQUIRES_T(( std::is_constructible<T, U>::value ))
-#endif
-    >
+    template< class U >
     gsl_api gsl_constexpr14 gsl_not_null_explicit
 #if gsl_HAVE( RVALUE_REFERENCE )
-    not_null( U && u )
+    not_null( U && u
+        gsl_REQUIRES_A((
+            // without constraining it, U&& would be too greedy,
+            // see https://mpark.github.io/programming/2014/06/07/beware-of-perfect-forwarding-constructors/
+            // alternative would be to have a "copy, then move" constructor
+            !detail::is_not_null_cvref<U>::value
+# if ! gsl_BETWEEN(gsl_COMPILER_CLANG_VERSION, 1, 400)
+            // in clang 3.x, is_constructible with T=unique_ptr<X>, U=not_null<T> tries to call copy constructor of unique_ptr, triggering an error
+            // it's okay to skip this check, because misuse will still trigger an error, but a less readable one
+            && std::is_constructible<T, U>::value
+# endif
+    )))
     : ptr_( std::forward<U>( u ) )
 #else
     not_null( U const & u )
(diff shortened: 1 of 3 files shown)


test/not_null.t.cpp
```diff
@@ -392,10 +799,10 @@ struct ETLessWeirdPtr // // element_type and type T differ
 using std::is_same;
 #else
 template<class T, class U>
-struct is_same { enum { value = false }; };
+struct is_same { enum dummy { value = false }; };
```
mbeutel: Why introduce a name for the enum?

test/not_null.t.cpp
```diff
@@ -20,6 +20,24 @@
...
+using std::make_shared;
+#endif
+
+#if gsl_HAVE( UNIQUE_PTR )
+using std::unique_ptr;
+# if gsl_HAVE( MAKE_UNIQUE )
+using std::make_unique;
+# else
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args)
```
mbeutel: Given that `make_unique<>()` was added only in C++14, perhaps it'd be a useful thing to have in `namespace gsl`, right next to the `unique_ptr<>` alias.

petamas:
> * I think `gsl_REQUIRES_A()` should go in the template argument list, i.e.
>   ```c++
>   template< class U
>       gsl_REQUIRES_A( ... ) >
>   ```
>   The only reason I can think of for putting it into a function signature would be if a compiler had trouble with SFINAE in the template argument list (or if `! gsl_HAVE( DEFAULT_FUNCTION_TEMPLATE_ARG )`, but when dealing with rvalue references or forwarding references we're in C++11 anyway).

Now you got me confused. In my understanding, std::enable_if can go to three places:
- as an extra function parameter (this is what `gsl_REQUIRES_A` is for)
- as an extra template parameter (this is what `gsl_REQUIRES_T` is for)
- as the return value (this is what `gsl_REQUIRES_R` is for)
In conclusion, I cannot move `gsl_REQUIRES_A` to the template parameter list without changing it to `gsl_REQUIRES_T`. However, in issue #178 you yourself say we should replace `gsl_REQUIRES_T` with `gsl_REQUIRES_A`, which is the reason I did it this way in my PR. If you want, I can change it to `gsl_REQUIRES_T` (I prefer that form anyway), I just wanted to be in line with your future plans. :)

Also note that there's at least two compilers without `DEFAULT_FUNCTION_TEMPLATE_ARG`, but with rvalue references, VS2010 (a.k.a. gsl_CPP11_100) and VS2012 (a.k.a. gsl_CPP11_110). Even though VS2010 does not have TYPE_TRAITS, VS2012 does, so we would lose move support if we changed back to gsl_REQUIRES_T. (Note that this is a loss only w.r.t. my PR, not v0.34.0)

> * I'd prefer a simple value-accepting constructor (`not_null(U u)`) over constrained forwarding references; I assume this is what you mean by "copy, then move". Note that it should never induce extra copies, only an extra move (except in the pathological case where a type is copyable but not movable). This change should vastly simplify the constructors and assignment operators of `not_null<>`.

Yes, that's exactly what I mean by "copy, then move", as per the linked blog post. I also like that version more, so I'll update my PR. (I chose U&& because I wanted to be as close as possible to the original version.) However, I don't believe this would help simplifying the other consturctors.

> * Forwarding references rely on reference collapsing, i.e. for `U &&` the compiler can infer `U = V&` and then collapse `U & &&` to `U &`. Note that, for this reason, `not_null<U> &&` is _not_ a forwarding reference but always an rvalue reference (and perhaps not what you intended).

Forwarding references were not my intent in the `not_null<U> &&` constructor - I wanted to do a "converting move consturctor", which takes an rvalue reference to a different type of not_null object (eg. not_null<unique_ptr<X>> for not_null<shared_ptr<X>>), and moves its contents to the internal T. 

> To summarize, wouldn't it be sufficient to do approximately this?
> 
> ```c++
> template< class U
>     gsl_REQUIRES_A( std::is_convertible<U, T>::value ) >
> explicit not_null(U u) { ... }
> template< class U
>     gsl_REQUIRES_A( std::is_convertible<U, T>::value ) >
> /* implicit*/ not_null(not_null<U> u) { ... }
> ```

I do not think that is enough. There are five cases we need to handle:
- some U type that is not a not_null specialization
- `not_null<U>` rvalue, where U is implicitly convertible to T - we want implicit conversion in this case
- `not_null<U>` lvalue, where U is implicitly convertible to T - we want implicit conversion in this case
- `not_null<U>` rvalue, where U is explicitly (but not implicitly) convertible to T - we want only explicit conversion in this case
- `not_null<U>` lvalue, where U is explicitly (but not implicitly) convertible to T - we want implicit conversion in this case

Your version handles the first three cases,  but it doesn't provide a solution to the 4th and 5th. Naturally, another (explicit) constructor like it (with a different condition) takes care of the problem, so we're up to 3 constructors. If we were targeting only C++11 (or newer), we could stop there, but since we want to target C++03 too, we need to provide a version for that too. (The non-reference version would do two copies, which is not something we want.) In summary, we will end up with 4 constructors - which is still one less than my solution, and the code will be much clearer, so I guess you're right.

Regarding assignment operators: during development, I had some weird build errors without explicitly spelling them out, but it seems those're fixed now, so I'll remove these operators from the C++11 part of the code. (Because all constructors are made explicit in C++03, we'll still need operator=() for those compilers.)

I'll react inline to your other comments.

include/gsl/gsl-lite.hpp
```diff
@@ -1372,26 +1433,75 @@ class not_null
...
+    gsl_api gsl_constexpr explicit not_null( not_null<U> const & other
+#if gsl_HAVE( TYPE_TRAITS )
+        gsl_REQUIRES_A((std::is_constructible<T, U>::value && !std::is_convertible<U, T>::value))
+#endif
+    )
+    : ptr_( other.checked_ptr_ref() )
+    {}
+
+    template< class U >
+    gsl_api gsl_constexpr14 not_null<T>& operator=( not_null<U> const & other)
```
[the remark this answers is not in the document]
petamas: I once had some problems with a compiler (I think qcc 4.7.3, a fork of gcc for the QNX platform) with this, so I'd prefer to leave it this way. Also, I think it's clearer this way, since there's already another not_null in this line.

test/not_null.t.cpp
petamas: It fixed -Wunnamed-type-template-args warnings introduced in PR #192

test/not_null.t.cpp
petamas: I think it would be better to place it in the gsl::std14 namespace first, then import it to gsl next to std::unique_ptr. Also, I did not notice this part of the header, so I guess I have some unneeded usings in my test. :)

petamas:
Please note that AppVeyor failed for this PR - I made a mistake VS2013 did not like. Please do not merge back until it is fixed (even if I update the PR with the other changes).

include/gsl/gsl-lite.hpp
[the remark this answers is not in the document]
mbeutel: OK.

mbeutel:
> Now you got me confused. In my understanding, std::enable_if can go to three places:
> 
>     * as an extra function parameter (this is what `gsl_REQUIRES_A` is for)

This may have been the intent when the macro was introduced (the name suggests this). But `gsl_REQUIRES_A()` is actually fine in template argument lists, and generally preferable to `gsl_REQUIRES_T()` for the reason explained here:
https://en.cppreference.com/w/cpp/types/enable_if (under "Notes"). (Also, `gsl_REQUIRES_T()` can appearently lead to `-Wunnamed-type-template-args` as you've discovered. The warning is bogus in C++11, but still.)

I'd like to apply these guidelines for SFINAE:

- If possible, put constraints on the return type (i.e. `gsl_REQUIRES_R()`). This form has the benefit that the function signature is not affected by the presence of SFINAE. Also, there is [anecdotal evidence](https://twitter.com/ericniebler/status/958490446107361280) that it compiles substantially faster.
- If this is not possible (e.g. in constructors), use the non-type template parameter form of SFINAE, i.e. `gsl_REQUIRES_A()` in the template argument list.
- Use `gsl_REQUIRES_A()` in the function signature as a fallback only for compilers known to have trouble with default function template arguments.


> I do not think that is enough. There are five cases we need to handle:
> 
> [...]
> 
> Your version handles the first three cases, but it doesn't provide a solution to the 4th and 5th.
>
> If we were targeting only C++11 (or newer), we could stop there, but since we want to target C++03 too, we need to provide a version for that too.

Agreed.

mbeutel:
> Please note that AppVeyor failed for this PR - I made a mistake VS2013 did not like. Please do not merge back until it is fixed (even if I update the PR with the other changes).

Yes, I learned that lesson the hard way :) I'll let CI to run through before merging anything.

petamas:
@mbeutel: Huh, I did not know that about gsl_REQUIRES_A, but I see your point now. I'll modify it to have int and 0, and move it to the template parameter list. (Behind a DEFAULT_FUNCTION_TEMPLATE_ARG check.)

petamas:
@mbeutel: Do you know if Odin Holmes' ACCU talk he's referring to in the Twotter thread is available somewhere online? I can't seem to find it.

mbeutel:
> @mbeutel: Do you know if Odin Holmes' ACCU talk he's referring to in the Twotter thread is available somewhere online? I can't seem to find it.

No, sorry. Perhaps just drop him a line? I'd be interested too.

petamas:
@mbeutel : I've updated the PR with the following:

- Change constructor parameters to U instead of U&& and const U&
- Fix semi-old Visual Studio compilation (has rvalue semantics, but does not have `= default`)
- Move make_unique to gsl-lite.hpp

Plan to do before merging:
- Move gsl_REQUIRES_A to template parameter list
- Simplify checked_ptr_ref() and checked_ptr_move() a little bit, making their use clearer

Also, I'll contact Odin about the talk. :)

petamas:
Seems like I messed up with an even older VS. Dangit. I'll fix this in my next update, but this time, I'll wait for AppVeyor to finish before I update the PR. :D

petamas:
@mbeutel: I think I'm finished with this PR, please check if I missed anything important. :)

include/gsl/gsl-lite.hpp
```diff
@@ -240,7 +240,7 @@
     typename std::enable_if<VA, R>::type
 
 #define gsl_REQUIRES_A(VA) \
-    , typename std::enable_if<VA, void*>::type = nullptr
+    , typename std::enable_if<VA, int>::type = 0
```
mbeutel: This is a potentially ABI-breaking change, but `gsl_REQUIRES_A()` is neither a documented feature of *gsl-lite* nor used in any class template argument lists, so I think we can take the risk.

include/gsl/gsl-lite.hpp
```diff
@@ -555,6 +577,48 @@ typedef integral_constant< bool, false > false_type;
...
+
+# elif gsl_HAVE( VARIADIC_TEMPLATE )
+
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args)
+{
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+
+# else
```
mbeutel: The following code (`unique_ptr<>` and rvalue references but no variadic templates) is exclusively for VS 2010/2012, right? I'd rather not support `make_unique<>()` there instead of starting to emulate variadic templates.

include/gsl/gsl-lite.hpp
```diff
@@ -1367,31 +1453,70 @@ class not_null
...
+    gsl_api gsl_constexpr14 not_null<T>& operator=( const not_null<U>& other)
+    {
+        ptr_ = other.checked_ptr();
+        return *this;
+    }
+#endif // gsl_HAVE( RVALUE_REFERENCE )
 
     gsl_api gsl_constexpr14 get_result_t get() const
     {
-        // Without cheating and changing ptr_ from the outside, this check is superfluous:
```
mbeutel: This comment was wrong (`ptr_` can legally be `nullptr` after the pointer was moved-from) so it's a good thing you took it out.

mbeutel:
Thank you, this is excellent work.

Would you mind taking out the non-variadic flavor of `make_unique<>()` (cf. comment above)? Apart from that everything looks good.

include/gsl/gsl-lite.hpp
petamas: I agree, the gsl_REQUIRES macros are implementation details, they are not meant to be used by consumers.

include/gsl/gsl-lite.hpp
petamas: Yes, it's only those compilers. I can take that branch out of the library, but since I use the functions for clarity in the unit tests, I need to add them to not_null.t.cpp. Is that okay?

mbeutel: Sure.

petamas: Will do, thanks.

Issue #185, opened 2019-09-18
Consider changing the return value type of not_null::get()


petamas:
Currently, there are two possible types for the return value of not_null::get():
```
#if gsl_CONFIG( NOT_NULL_GET_BY_CONST_REF )
    typedef T const & get_result_t;
#else
    typedef T get_result_t;
#endif
```

However, I believe both to be wrong.

Simply using `T` does not play well with unique_ptr (see issue #129), but `const T&` is also not that comfortable: if `T` is a smart pointer, and someone would like to get the raw pointer behind it, they need to write `p.get().get()`, which is ugly.

I believe `get()` should return an `element_type*`, where `element_type` is `remove_ref_t<*declval<T>()>`, and `get()` should use the smart pointer's `get()` method. (Of course, this would need a specialization for raw pointers, but that can be done easily.) I believe the other operators (eg. * or ->) should also be defined using `element_type`.

To access the underlying (maybe smart) pointer within `not_null` (same semantics as current `get()`), I propose an `as_nullable()` method. If you do not want to change the semantics of `get()` (by returning the raw pointer instead of the underlying pointer), I propose a `get_raw()` method with an `element_type*` return type.

(I got these ideas by comparing gsl::not_null with Dropbox nn: https://github.com/dropbox/nn)

petamas:
@mbeutel: your issue #179 seems to be related to this.

mbeutel:
The [Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#i12-declare-a-pointer-that-must-not-be-null-as-not_null) don't specify exactly what `not_null<>` is supposed to do, but my understanding is that it is meant for "stating the intent in source". So the following would be a conforming implementation:

```c++
template <typename P> using not_null = P;
```

If my reading is correct, anything beyond that (like the actual runtime checks we do) is just QoI. But then clearly `not_null<P>::get()` should return `decltype(P::get())`, not `P`.

I think your alternative 1 is the way to go, though I'd wait for version 1.0 because this is a breaking change.

petamas:
I agree with your interpretation of the guidelines.

I think we could put "alternative 1" behind a configuration macro for now, then make that the default behaviour in 1.0. What do you think?

mbeutel:
It should be noted that this change would alter the meaning of the following code:
```c++
std::shared_ptr<int> foo(gsl::not_null<std::shared_ptr<int>> p)
{
[...]

PR #205, opened 2019-10-10
Fix awkward issues in `not_null<>` introduced in #202; add tests

mbeutel:
- `get()` should not need function ref qualifiers
- `get()` should return a plain pointer
- `operator->()` can safely return `T const &` in all cases
- `operator*()` should always return `element_type &`

Thanks to @petamas for feedback.


test/not_null.t.cpp
```diff
@@ -443,7 +482,26 @@ CASE( "not_null<>: Allows to construct from a non-null underlying pointer (uniqu
...
+    EXPECT( pg == raw );
+#elif gsl_CONFIG( NOT_NULL_GET_BY_CONST_REF )
+    int* pg = p.get().get();
+    EXPECT( pg == raw );
+#else
+    (void) raw;
+    (void) pi;
+    (void) p;
+#endif
+}
```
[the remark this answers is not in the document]
mbeutel: > I still think "transparent" and "constref" should be mutually exclusive, and "transparent" should imply the parts of "constref" that are not the return value of get().

So you're saying `gsl_CONFIG_NOT_NULL_TRANSPARENT_GET=1` should imply `gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF=1`, but explicitly defining `gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF` to any value would then be an error?

include/gsl/gsl-lite.hpp
```diff
@@ -1540,32 +1540,22 @@ class not_null
...
-    gsl_api gsl_constexpr14 element_type && operator*() &&      { return *std::move(checked_ptr()); }
-
 #else
-    gsl_api gsl_constexpr14 get_result_t get() const { return checked_ptr(); }
-
     gsl_api gsl_constexpr   operator get_result_t  () const   { return checked_ptr(); }
+#endif
 
-    gsl_api gsl_constexpr   get_result_t operator->() const   { return checked_ptr(); }
+    gsl_api gsl_constexpr   T const &     operator->() const   { return checked_ptr(); }
```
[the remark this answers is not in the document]
mbeutel: I believe an `operator->()` returning anything other than `T const &` is fundamentally broken. Consider what would happen if you had a `not_null<auto_ptr<T>> p` and then did `p->f()`.

test/not_null.t.cpp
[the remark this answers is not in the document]
mbeutel: Agreed.

include/gsl/gsl-lite.hpp
```diff
@@ -1540,32 +1540,22 @@ class not_null
...
-# else
-    gsl_api gsl_constexpr14 get_result_t get() const { return checked_ptr(); }
-# endif
+#if gsl_CONFIG( NOT_NULL_TRANSPARENT_GET )
+    gsl_api gsl_constexpr14 element_type* get() const { return checked_ptr().get(); }
+#else
+    gsl_api gsl_constexpr14 get_result_t  get() const { return checked_ptr(); }
+#endif
 
+#if gsl_HAVE( FUNCTION_REF_QUALIFIER )
```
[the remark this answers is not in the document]
mbeutel: I'd be very unhappy with that because then our default configuration wouldn't support moving from `not_null<unique_ptr<T>>`.

Ideally, we could have
```c++
    gsl_api gsl_constexpr14 operator T const &() const & { return checked_ptr(); }
    gsl_api gsl_constexpr14 operator T &&     () &&      { return std::move(checked_ptr()); }
```

so we would avoid a copy but still permit a move. However, this makes the following code ambiguous:

```c++
not_null<unique_ptr<T>> np = ...;
unique_ptr<T> p = std::move(np);
```

However, I think having the implicit conversion operator return a value is justifiable in C++11 because move semantics alleviate the cost of moving the returned value around.

The only case where converting to `T const &` would save you a copy is when interacting with code like this, which I'd hope to be rare:

```c++
// Maybe we want a copy, maybe we don't, so take the arg by const ref.
// (If we knew we needed a copy, we'd simply take the arg by value;
//  if we only wanted to borrow the pointer we'd just take a plain `int const *` instead.)
void foo( std::shared_ptr<int> const & );
```

I think not supporting moving from non-copyable smart pointers is too high a price to pay for this.

mbeutel:
Commit https://github.com/martinmoene/gsl-lite/pull/205/commits/fcb1c92793087966e67149d964079e389bb2ff62 makes our conversion logic consistent with the construction logic so we can also convert to related pointer types. I also added some comments explaining the design choice of converting to `T` rather than `T const &`.

petamas:
(I just want to tell you I won't be able to answer today, but I'll get back to you tomorrow.)

mbeutel:
> (I just want to tell you I won't be able to answer today, but I'll get back to you tomorrow.)

Would you be ok with the current state? If so, I'd like to merge it.

include/gsl/gsl-lite.hpp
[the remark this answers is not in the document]
petamas: It is fundamentally broken, but I think the whole gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF=0 part was broken :D

include/gsl/gsl-lite.hpp
```diff
@@ -1540,32 +1560,54 @@ class not_null
...
+    gsl_api gsl_constexpr14 explicit operator U gsl_not_null_CONVERSION_REF() const gsl_not_null_LVALUE_REF { return checked_ptr(); }
+# if gsl_HAVE( FUNCTION_REF_QUALIFIER )
+    template< class U
+        gsl_REQUIRES_A(( std::is_constructible<U, T>::value && !std::is_convertible<T, U>::value && !gsl::detail::is_not_null_oracle<U>::value ))
+    >
+    gsl_api gsl_constexpr14 explicit operator U() && { return std::move(checked_ptr()); }
+# endif
 
-    gsl_api gsl_constexpr   operator get_result_t  () const   { return checked_ptr(); }
+    // implicit conversion operator if type_traits is available
```
petamas: This comment is at least true, but it was only used originally as a contrast with the previous comment. I don't think it has a function here.

include/gsl/gsl-lite.hpp
```diff
@@ -1540,32 +1560,54 @@ class not_null
...
+    gsl_api gsl_constexpr14 get_result_t  get() const { return checked_ptr(); }
+#endif
 
-    gsl_api gsl_constexpr14 T const & operator->() const & { return checked_ptr(); }
-    gsl_api gsl_constexpr14 T &&      operator->() &&      { return std::move(checked_ptr()); }
+    // Ideally, we would convert to either `T const &` or `T &&` so we would permit a move but still avoid potentially unnecessary copies.
+    // Unfortunately, that would render code like `std::unique_ptr<T> p = not_null<std::unique_ptr<T>>(...);` ambiguous.
+    // We choose to always convert to `T` instead if function ref qualifiers are available, expecting that the cost of the copy will usually be
+    // alleviated by move semantics.
+    // This implies we cannot convert to const refs of move-only types; but that scenario should be rare.
```
petamas: I have a hard time understanding this comment. Especially this part:

> We choose to always convert to `T` instead if function ref qualifiers are available

However, we're NOT converting to T (as in the parameter of not_null), but U (a type T is convertible to). If I interpret this correctly, you used "T" here as a wildcard type, but it's confusing because T is a name already used for something else in this context.

test/not_null.t.cpp
```diff
@@ -443,7 +541,24 @@ CASE( "not_null<>: Allows to construct from a non-null underlying pointer (uniqu
     EXPECT( &*p == raw );
 }
 
-CASE( "not_null<>: Allows to move from a not-null pointer (unique_ptr)" )
+CASE( "not_null<>: Returns underlying pointer with get() (unique_ptr)" )
```
petamas: This test case is not named well: when "transparent get" is on, we're not returning the underlying pointer (I've used the term "underlying pointer" to refer to T previously).

test/not_null.t.cpp
```diff
@@ -274,6 +326,34 @@ CASE( "not_null<>: Allows to construct from a non-null underlying pointer (share
...
+    shared_ptr< int > pi = make_shared< int >(12);
+    int* raw(pi.get());
+
+    not_null< shared_ptr< int > > p ( std::move(pi) ); // There...
+    pi = std::move(p); // ...and back again.
+
+    EXPECT_THROWS( *p );
+    EXPECT( pi.get() == raw );
+#endif
+}
```
petamas: Why is this test only supported when FUNCTION_REF_QUALIFIER is supported? This should work on gcc 4.7.3 too!

Also, the test case's name is misleading: it does not mention that we're moving to an underlying pointer.

include/gsl/gsl-lite.hpp
```diff
@@ -1439,6 +1448,17 @@ class not_null
...
+# define gsl_not_null_LVALUE_REF &
+# define gsl_not_null_CONVERSION_REF
+#else
+# define gsl_not_null_LVALUE_REF
+# if gsl_CONFIG( NOT_NULL_GET_BY_CONST_REF )
+#  define gsl_not_null_CONVERSION_REF const &
+# else
+#  define gsl_not_null_CONVERSION_REF
+# endif
+#endif
```
petamas: These should be moved closer to their uses to make understanding the code easier.

test/not_null.t.cpp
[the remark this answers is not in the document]
petamas: `So you're saying gsl_CONFIG_NOT_NULL_TRANSPARENT_GET=1 should imply gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF=1, but explicitly defining gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF to any value would then be an error?`

In a sense, yes. But it would be more precise to say I think get's return value and the operators' return value should be separated from each other, because when "get is transparent", then "get is returning const ref" is not true, but I think we want the conversion and arrow operators to behave in a more modern way in both cases.

include/gsl/gsl-lite.hpp
mbeutel: So should `gsl_not_null_explicit`, I suppose? I thought I'd keep the macro definitions together as a group... but yes, I can move them closer.

test/not_null.t.cpp
mbeutel: I fail to see how this could work without `FUNCTION_REF_QUALIFIER`. The conversion operator plainly fails for `not_null<unique_ptr<T>>` because it cannot call `move()`.

include/gsl/gsl-lite.hpp
```diff
@@ -1540,32 +1560,54 @@ class not_null
...
+    // Unfortunately, that would render code like `std::unique_ptr<T> p = not_null<std::unique_ptr<T>>(...);` ambiguous.
+    // We choose to always convert to `T` instead if function ref qualifiers are available, expecting that the cost of the copy will usually be
+    // alleviated by move semantics.
+    // This implies we cannot convert to const refs of move-only types; but that scenario should be rare.
 
-    gsl_api gsl_constexpr14 element_type &  operator*() const & { return *checked_ptr(); }
-    gsl_api gsl_constexpr14 element_type && operator*() &&      { return *std::move(checked_ptr()); }
+#if gsl_HAVE( RVALUE_REFERENCE ) && gsl_HAVE( TYPE_TRAITS ) && gsl_HAVE( DEFAULT_FUNCTION_TEMPLATE_ARG ) && gsl_HAVE( EXPLICIT )
+    // explicit conversion operator
+    // if type_traits is not available, then we can't distinguish is_convertible and is_constructible, so we use this unconditionally
```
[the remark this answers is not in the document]
mbeutel: Right, that was a leftover from the last commit when I realized VS 2012 wouldn't be able to handle any of the SFINAE anyway and thus moved the weaker condition `gsl_HAVE( TYPE_TRAITS )` upwards too.

include/gsl/gsl-lite.hpp
mbeutel: Well, it *does* summarize the meaning of the 9 following unreadable lines quite well.

include/gsl/gsl-lite.hpp
mbeutel: I'll try to rephrase.

```c++
// We want an implicit conversion operator that can be used to convert from both lvalues (by
// const reference or by copy) and rvalues (by move). So it seems like we could define
//     template< class U, gsl_REQUIRE_A(...) >
//     operator U const &() const & { return checked_ptr(); }
//     template< class U, gsl_REQUIRE_A(...) >
//     operator U&&() && { return std::move( checked_ptr() ); }
// However, having two conversion operators with different return types renders the assignment
// operator of the result type ambiguous:
//     not_null<unique_ptr<T>> p( ... );
//     unique_ptr<U> q;
//     q = std::move( p ); // ambiguous
// To avoid this ambiguity, we have both overloads of the conversion operator return `U`
// rather than `U const &` or `U &&`. This implies that converting an lvalue always induces
// a copy, which can cause unnecessary copies or even fail to compile in some situations:
//     not_null<shared_ptr<T>> sp( ... );
//     shared_ptr<U> const & rs = sp; // unnecessary copy
//     unique_ptr<U> const & ru = p; // error: cannot copy `unique_ptr<T>`
// However, these situations are rather unusual, and the following, more frequent situations
// remain unimpaired:
//     shared_ptr<U> vs = sp; // no extra copy
//     unique_ptr<U> vu = std::move( p );
```

Is that better? Also much longer, though.

test/not_null.t.cpp
mbeutel: Would "Returns underlying pointer or raw pointer with get()" be better?

mbeutel:
> TBH, I'm not a fan of this changeset as a whole: the conversion code is basically unreadable

That is certainly true; but to be fair, the conversion constructors are also a little hard to parse. In a sense I only re-established symmetry there :)

More seriously, though: ugliness is the nature of generic conversion operators. Personally I don't have much love for `not_null<>`; whenever I wanted to use it, I figured I could instead redesign the code to avoid pointer semantics, and hence the problem of nullability, entirely.


> and I have a feeling that gcc 4.7.3 (no func. ref. qualifiers) will not behave as expected.

It will eagerly permit conversion and then fail during instantiation, which can break other code in subtle ways (e.g. SFINAE checks that ask for convertibility). And I think it cannot properly support conversions from rvalues (see above). Is there anything else you are thinking of?

test/not_null.t.cpp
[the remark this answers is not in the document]
mbeutel: Yes, I think we're on the same page now. Please see my latest commit which renders `gsl_CONFIG_NOT_NULL_TRANSPARENT_GET` and `gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF` mutually exclusive.

include/gsl/gsl-lite.hpp
petamas: Much, much better. :)

mbeutel:
Great, then I'm merging now. Thanks for your thorough review!

> Also, the remaining problems with the current approach - like the not compiling examples in your comment, or all the conversions w/ gcc 4.7.3 - can be alleviated by the use of a properly defined as_nullable().

A free function `as_nullable()` is a suitable workaround for the lack of function-level ref modifiers in GCC 4.7. But `as_nullable()` should also return a value, not a reference, otherwise we run into the same ambiguity; and hence it cannot make the two ref-binding examples compile either. We could implement an additional `as_nullable_ref()` if we have to, but I'd rather not.

A PR for `as_nullable()` is welcome :) I'd like to release 0.36 before next week though; do you want this included in the release?

petamas:
I'd have liked to have it in the next release, yeah, but it is not feasible for me to do the PR this week, and I'd really want to do it myself (there are a few corner cases I want to play around with), so I guess I'll delay it until 0.37. (And if I'm under time constraints, I'll simply use 0.36 + my PR in the company codebase.)

> But `as_nullable()` should also return a value, not a reference, otherwise we run into the same ambiguity; and hence it cannot make the two ref-binding examples compile either.
I think if we overload in C++03 style, then we can circumvent these issues, but I'll need some rigorous testing to check it:
```
    gsl_api gsl_constexpr   T const & as_nullable() const { return checked_ptr(); }
    gsl_api                 T &       as_nullable()       { return checked_ptr(); }
```
Note that you'll need to use `std::move(np.as_nullable())` instead of `std::move(np).as_nullable()`, but I think it'll work. But I'll try it out, and add tests, as usual. :)

petamas:
> That is certainly true; but to be fair, the conversion constructors are also a little hard to parse. In a sense I only re-established symmetry there :)

Well, I think the conversion constructors are much more readable than your previous version for two reasons:
1. there are no syntax-changing macros (like gsl_not_null_LVALUE_REF) in them (this became much nicer in the final version)
2. I wrote the conversion constructors, so naturally, I can follow their logic more easily (also, they are by definition perfect :P)

> More seriously, though: ugliness is the nature of generic conversion operators. Personally I don't have much love for `not_null<>`; whenever I wanted to use it, I often figured I could instead redesign the code to avoid pointer semantics, and hence the problem of nullability, entirely.

My problem with them was more the macro magic, and less the "noise" of generic conversion operators.
Regarding `not_null<>` in general: for raw pointers, I agree whole-heartedly, but for shared ownership of a non-null resource, I haven't yet found a good alternative to `not_null<shared_ptr<T>>`.

mbeutel:
> Note that you'll need to use `std::move(np.as_nullable())` instead of `std::move(np).as_nullable()`

Note that, in either case, `as_nullable()` should be a free function, not a member function (cf. my draft at https://github.com/martinmoene/gsl-lite/issues/185#issuecomment-539607293).

The reason is that the Core Guidelines specify `not_null<>` as transparent, i.e.
```c++
template< class T> using not_null = T;
```
would be a conforming implementation. A no-op implementation of the GSL (let's call it "zGSL", as in "zero-overhead") that wants to be compatible with gsl-lite could then provide a simple pass-through implementation of `as_nullable()`.

Following similar reasoning (switching between `T` and `not_null<T>` shouldn't break code), I think we should also offer a pass-through overload of `as_nullable()` for nullable pointers as I did in my draft.

Note also that we want to suppress ADL for `as_nullable()` to keep users from getting away with unqualified calls that would break when switching to either "zGSL" or to the underlying pointer type. This can be done by hoisting the symbol from a nested namespace:

```c++
namespace gsl {
namespace detail {

template< class T >
auto as_nullable( not_null<T> const & p ) { ... };

} // namespace detail

using gsl::detail::as_nullable;

} // namespace gsl
```

petamas:
You're right, I haven't noticed that part of your comment. I understand your reasoning for `as_nullable` being a free function, I'll do it accordingly. (Maybe I'll add a deleted version as a member function, to help developers using autocomplete discover the feature.)

mbeutel:
> `not_null<>` cannot be used here without introducing an extra flag for the disposed state.

Actually I was wrong, and it can be used there just fine, and to great effect:

```c++
class File
{
private:
    struct FileDeleter
    {
        void operator ()(FILE* file) const noexcept
        {
                // We intentionally ignore failure of resource disposal here.
            std::fclose(file);
        }
    };

    gsl::not_null<std::unique_ptr<std::FILE, FileDeleter>> handle_;

public:
    File(std::filesystem::path const& path);

    void close()
    {
        FILE* rawHandle = gsl::as_nullable(std::move(handle_)).release();
        int ec = std::fclose(rawHandle);
        if (ec != 0) throw IOError(ec);
    }

    // file operations
};
```

You will also notice that I found a use case for `as_nullable()` here :) But it is needed here only because `not_null<>` isn't perfectly transparent. We could also add "forwarders" for typical smart pointer member functions such as `release()` and `reset()`, similar to what we did for `get()`.

I opened #207 to discuss the question of "forwarders".

PR #208, opened 2019-10-17
  Unify and document SFINAE conventions; add missing checks in not_null<> move and copy constructors

mbeutel:
* Unify and document our SFINAE conventions

* Add missing checks in `not_null<>` move and copy constructors

  We had non-null checks in converting constructors but not in plain
  move/copy constructors. We want to be consistent and have checks
  everywhere.

  Also, pre- and postconditions are now expressed with `Expects()` and
  `Ensures()` directly in the functions that impose them rather than
  being delegated to `checked_ptr()`. This is preferable because of the
  self-documenting nature of contract checks, and because it is closer to
  the Contracts draft language feature which makes contract checks part
  of the function signature.


mbeutel:
One side-effect of manually implementing move and copy assignment operator is that `std::is_copy_constructible<not_null<std::unique_ptr<T>>>` will wrongly answer `true` now. That ought to be fixed before merging this.

Issue #194, opened 2019-09-24
Collisions with M-GSL


mbeutel:
*gsl-lite* lives in the same namespace as Microsoft's GSL. This is reasonable because both are supposed to implement the Guidelines Support Library loosely specified in the Core Guidelines, and because we want to make it easy to switch between implementations by just swapping out `#include <gsl/gsl>` for `#include <gsl/gsl-lite.hpp>`.

However, this means that both libraries cannot coexist. One usually wouldn't want to use both at once; but if either is used as a public dependency of a library, this can cause conflicts for library consumers who use a different GSL in their own code.

At present this problem isn't likely to occur because only few libraries use the GSL; on [Vcpkg](https://github.com/microsoft/vcpkg) I could find only one library using *gsl-lite* ([telnetpp](https://github.com/microsoft/vcpkg/blob/master/ports/telnetpp)) and one using M-GSL ([coroutine](https://github.com/microsoft/vcpkg/blob/master/ports/coroutine))). But I think the GSL would be a useful ingredient for most libraries (if only for the precondition checking macros)., and I'd be happier if I could unhesitatingly recommend either of M-GSL and *gsl-lite* for use in libraries.

So I was thinking of the following plan:

- Move the implementation to a different header file, say, `gsl/gsl-lite-standalone.hpp` (though hopefully a better suggestion will come up).
- Rename the namespace to `gsl_lite`. This ensures that no link-time incompatibilities with eponymous types in M-GSL can occur.
- Rename the precondition checking macros to `gsl_EXPECTS()`, `gsl_ENSURES()` etc. (M-GSL prefixes its macros with upper-case `GSL_`, whereas `gsl_` is now idiomatic for *gsl-lite*.)
- Add back a header `gsl/gsl-lite.hpp` which does
    ```c++
    #if defined( GSL_GSL_H ) || defined( GSL_CONTRACTS_H ) || ...
    # error Cannot use <gsl/gsl-lite.hpp> header together with headers from Microsoft GSL.
    #endif
    namespace gsl = gsl_lite;
    #define Expects(x) gsl_EXPECTS(x)
    ...
    ```

Then we could recommend that libraries use *gsl-lite* in "standalone mode". Of course they could say `namespace gsl = gsl_lite` inside their own namespace so their code doesn't have to change (except for the contract-related macros).

Any thoughts on this?

petamas:
I like this direction (and mused about it recently), but have a few notes:

[...]

PR #221, opened 2019-12-27
Add dedicated <gsl-lite/gsl-lite.hpp> header for v1

mbeutel:
As a ~first~ second step towards M-GSL compatibility (cf. #194, #222), this PR proposes the following change:

- Add new header file \<gsl-lite/gsl-lite.hpp\> which currently just forwards to the old header \<gsl/gsl-lite.hpp\>. It raises an error if `!(gsl_CONFIG_DEFAULTS_VERSION >= 1)`.

The idea is that libraries which want to use *gsl-lite* should do the following:

```c++
#include <gsl-lite/gsl-lite.hpp> // instead of <gsl/gsl-lite.hpp>

namespace foo {

namespace gsl = ::gsl_lite; // convenience alias

double mean(gsl::span<double const> elements)
{
    gsl_Expects(!elements.empty()); // instead of Expects()
    ...
}

} // namespace foo
```

Subsequently, version 1 of *gsl-lite* would then no longer define the namespace `gsl` in the \<gsl-lite/gsl-lite.hpp\> header. Library "foo" will then happily coexist with other code which uses either M-GSL or *gsl-lite* in "traditional mode" (or "GSL mode", or whatever we want to call it) by defining `gsl_FEATURE_GSL_MODE=1`, or by including the old header \<gsl/gsl-lite.hpp\> . (Cf. the comment below for more details; I'd appreciate thoughts and comments especially on this part.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,10 @@ install(
     FILES       "include/gsl/gsl-lite.hpp"
     DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/gsl/"
 )
+install(
+    FILES       "include/gsl-lite/gsl-lite.hpp"
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/gsl-lite/"
+)
 if(GSL_LITE_OPT_INSTALL_COMPAT_HEADER)
     install(
         FILES       "include/gsl/gsl"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ cmake_minimum_required( VERSION 3.5 FATAL_ERROR )
 
 project(
     gsl_lite
-    VERSION 0.35.3
+    VERSION 0.35.4
 #   DESCRIPTION "A single-file header-only version of ISO C++ Guideline Support Library (GSL) for C++98, C++11 and later"
 #   HOMEPAGE_URL "https://github.com/martinmoene/gsl-lite"
     LANGUAGES CXX )
--- a/README.md
+++ b/README.md
@@ -302,10 +302,12 @@ Provide experimental types `final_action_return` and `final_action_error` and co
 
 There are four macros for expressing pre- and postconditions:
 
-- `Expects()` for simple preconditions
-- `Ensures()` for simple postconditions
+- `gsl_Expects()` for simple preconditions
+- `gsl_Ensures()` for simple postconditions
 - `gsl_ExpectsAudit()` for preconditions that are expensive or include potentially opaque function calls
 - `gsl_EnsuresAudit()` for postconditions that are expensive or include potentially opaque function calls
+
+The macros `Expects()` and `Ensures()` are also provided as aliases for `gsl_Expects()` and `gsl_Ensures()`.
 
 
 The following macros control whether contracts are checked at runtime:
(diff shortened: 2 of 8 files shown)


mbeutel:
I'm still not quite sure about that new \<gsl-lite/gsl-lite.hpp\> header file.

I feel that \<gsl-lite/gsl-lite.hpp\> would be the appropriate header name for *gsl-lite* if we want *gsl-lite* to be more than just another GSL implementation, just as `gsl-lite::gsl-lite` would be the appropriate name for the CMake target. But I'm not sure if the benefits of making these changes outweigh the drawbacks.

If we do want to proceed with a dedicated `gsl_lite` namespace, I see two possibilities for version 1:

- Do not introduce a new header file. There is only \<gsl/gsl-lite.hpp\>, and it defines the namespace `gsl_lite` by default, and additionally the namespace `gsl` if `gsl_FEATURE_GSL_MODE` is defined:  
    
  ```c++
  // gsl/gsl-lite.hpp
  namespace gsl_lite { ... }
  namespace gsl_compat { using gsl_lite::span; ... }
  
  // Including the new header <gsl-lite/gsl-lite.hpp> enables "GSL mode" only if requested
  // by a feature macro.
  #if defined( gsl_FEATURE_GSL_MODE )
  namespace gsl = gsl_compat;
  # define Expects( x ) gsl_Expects( x )
  # define Ensures( x ) gsl_Ensures( x )
  #endif // defined( gsl_FEATURE_GSL_MODE )
  ```
    
  **Advantages:** less complexity and confusion, still trivially single-header  
    
  **Disadvantages:** breaks old code which expects everything to be in namespace `gsl`; the fix is trivial (define `gsl_FEATURE_GSL_MODE` in the build system), but it still is a breaking change, and we cannot really generate helpful diagnostics

- Introduce a new header file \<gsl-lite/gsl-lite.hpp\> in addition to \<gsl/gsl-lite.hpp\>.

  ```c++
  // gsl-lite/gsl-lite.hpp
  namespace gsl_lite { ... }
  namespace gsl_compat { using gsl_lite::span; ... }
  
  // Including the new header <gsl-lite/gsl-lite.hpp> enables "GSL mode" only if requested
  // by a feature macro.
  #if defined( gsl_FEATURE_GSL_MODE )
  namespace gsl = gsl_compat;
  # define Expects( x ) gsl_Expects( x )
  # define Ensures( x ) gsl_Ensures( x )
  #endif // defined( gsl_FEATURE_GSL_MODE )
  ```

  ```c++
  // gsl/gsl-lite.hpp
  #include <gsl-lite/gsl-lite.hpp>
  
  // Including the traditional header <gsl/gsl-lite.hpp> implicitly requests "GSL mode".
  //
  // If `gsl_FEATURE_GSL_MODE` is defined, the "GSL mode" definitions are already provided
  // by <gsl-lite/gsl_lite.hpp>, but they are idempotent, hence we can just state them again.
  namespace gsl = gsl_compat;
  # define Expects( x ) gsl_Expects( x )
  # define Ensures( x ) gsl_Ensures( x )
  ```
    
  **Advantages:** old code which references \<gsl/gsl-lite.hpp\> and expects a namespace `gsl` keeps working; \<gsl-lite/gsl-lite.hpp\> can still be used as single-header library (although then the `gsl_FEATURE_GSL_MODE` feature macro is required to enable the traditional mode)
    
  **Disadvantages:** more headers and more complexity; the distinction between  \<gsl/gsl-lite.hpp\> and  \<gsl-lite/gsl-lite.hpp\> is perhaps too subtle

@martinmoene and @petamas, I'd be grateful if you could take the time to comment on these thoughts. Of course this can wait until next year :)

martinmoene:
@mbeutel noticed your request :)

mbeutel:
The new `gsl_lite` namespace, probably less controversial, is now tracked separately in #222.

mbeutel:
I've been thinking about this for a while, and I believe that the new header file is the superior choice. Those who need a single-header GSL implementation will then need to define a build configuration macro like `gsl_FEATURE_GSL_MODE`, which seems bearable.

martinmoene:
@mbeutel sorry for slacking, couldn't make up my mind, hadn't come further than the split you made to enable different time paths. Support the choice you made, whichever it would have been. 

Like to mention I'm very happy and grateful for the professional approach and the direction you take *gsl lite*. Thanks very much!

mbeutel:
No worries! Thank you for your support :)

PR #222, opened 2019-12-30
Add dedicated `gsl_lite` namespace for v1

mbeutel:
As a first step towards M-GSL compatibility (cf. #194), this PR makes the following changes:

- If `gsl_CONFIG_DEFAULTS_VERSION >= 1`, introduce a new namespace `gsl_lite`. Right now it consists of a bunch of `using` statements pulling in most symbols from namespace `gsl`.
- Rename `Expects()` and `Ensures()` to `gsl_Expects()` and `gsl_Ensures()`. The old names are still provided as aliases.

The idea is that libraries which want to use *gsl-lite* should do the following:

```c++
#include <gsl/gsl-lite.hpp>

namespace foo {

namespace gsl = ::gsl_lite; // convenience alias

double mean(gsl::span<double const> elements)
{
    gsl_Expects(!elements.empty()); // instead of Expects()
    ...
}

} // namespace foo
```

Because we now have our own namespace, any extensions that are not in the Core Guidelines should go there. As a first example, this PR also adds three new typedefs `gsl_lite::dim`, `gsl_lite::stride`, and `gsl_lite::diff` which are aliases to the same type as `gsl::index` but permit more fine-grained semantics:

```c++
// assuming `namespace gsl = ::gsl_lite;`

// Compute matrix–vector product yᵢ = Σⱼ Aᵢⱼxⱼ
gsl::dim nx = gsl::ssize( x ),
         ny = gsl::ssize( y );
gsl_Expects( gsl::ssize( A ) == nx*ny );
gsl::stride sx = 1; // assume row-major order
            sy = nx;
std::fill( y.begin(), y.end(), 0. );
for ( gsl::index i = 0; i < ny; ++i ) {
    for ( gsl::index j = 0; j < nx; ++j ) {
        y[ i ] += A[ sy*i + sx*j ] * x[ j ];
    }
}
```

The proposal to add a new header \<gsl-lite/gsl-lite.hpp\> is tracked separately in #221.


mbeutel:
Some possible alternatives to `gsl_lite`:

- We could instead occupy the nested namespace `gsl::lite`. I don't like this too much because we're still in the same namespace as M-GSL.

- Or we could choose `nongsl`, similar to Martin's other "*-lite" libraries which live in namespace `nonstd`.

martinmoene:
Perhaps unsurprisingly I like namespace name `nongsl`. However it isn't shared over multiple libraries like `nonstd` is, and hence it might be a bit surprising to deviate from the name *gsl lite* :)

mbeutel:
Yeah, I think `gsl_lite` is the best choice for now. Merging.

PR #249, opened 2020-05-20
Fix compile issue due to \x115F unicode characters in comments

Maximus5:
The problem is that `gsl-lite.hpp` contains U+115F characters, but has no BOM.
And Visual Studio shows warning during compilation

    warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss

I believe these glyphs appeared in comments unintentionally.


mbeutel:
Sorry for answering late.

These Unicode characters were placed intentionally; cf. https://github.com/mbeutel/makeshift/issues/5 for a similar issue. But I'll happily get rid of them if that helps avoid encoding issues.

PR #251, opened 2020-07-25
Add as_nullable(not_null) function (w/ tests)

petamas:


mbeutel:
Thank you for the PR.

I didn't review the code yet, but let me bring up a concern regarding ADL which I had voiced in an old PR:

> [...] the Core Guidelines specify `not_null<>` as transparent, i.e.
> ```c++
> template< class T > using not_null = T;
> ```
> would be a conforming implementation. A no-op implementation of the GSL (let's call it "zGSL", as in "zero-overhead") that > wants to be compatible with gsl-lite could then provide a simple pass-through implementation of `as_nullable()`.
> 
> [...]
>
> Note also that we want to suppress ADL for `as_nullable()` to keep users from getting away with unqualified calls that would break when switching to either "zGSL" or to the underlying pointer type. This can be done by hoisting the symbol from a nested namespace:
>
> ```c++
> namespace gsl {
> namespace detail {
> 
> template< class T >
> auto as_nullable( not_null<T> const & p ) { ... };
> 
> } // namespace detail
> 
> using gsl::detail::as_nullable;
> 
> } // namespace gsl
> ```

_Originally posted by @mbeutel in https://github.com/gsl-lite/gsl-lite/pull/205#issuecomment-542696260_

petamas:
Right, I wanted to ask you about that last year, but forgot. To be honest, I don't understand that comment at all:
- What do you mean by "wants to be compatible with gsl-lite"? Compatible in what way? By adding an as_nullable (which is a nonstandard extension), or...? Also, please note that my implementation has overloads handling nullable types. (Inspired by your `make_not_null`, which doesn't re-wrap a `not_null`)
- I don't see why the unqualified calls breaking would be that problematic; if the user wants, they can `using gsl::as_nullable` to make their code compile again.

So, could you please elaborate on your points, maybe with a more detailed example (how would zGSL and the client code look, and what kind of transition would break it). I'm not strongly against disabling ADL, but I'd like to understand why I'm doing it. :) (Also, I use unqualified `as_nullable`  in my codebase by conscious decision, so I'll need to add the `using` statement, but that's not a big problem for me, since I already have a `not_null_extensions.h` with typedefs like `nn_unique_ptr` in my namespace.)

mbeutel:
Example:

```c++
// user code
void bar(X x);
void foo(gsl::not_null<X> nnx)
{
    auto x = as_nullable(nnx)); // finds gsl::as_nullable() via ADL
    bar(x);
}
```

```c++
// hypothetical gsl-lite-compatible zero-overhead GSL implementation
namespace zgsl {

template <typename T> using not_null = T;

template <typename T>
T as_nullable(T obj)
{
    return obj;
}

} // namespace zgsl
```

Now assume the user switches over his code to "zGSL", either by putting "namespace gsl = zgsl;" above the source file or by searching and replacing `gsl` -> `zgsl`. Then the code no longer compiles. To prevent this, I want the code to fail in the first place, which is achieved by suppressing ADL.


> Also, please note that my implementation has overloads handling nullable types.

Noted :)


> * I don't see why the unqualified calls breaking would be that problematic; if the user wants, they can `using gsl::as_nullable` to make their code compile again.

It's not a big deal, only a stumbling block that can be avoided.

---

It is my perception that making generic helper functions like `std::size()` discoverable via ADL has been widely regarded as a bad move. The helper functions added by the C++20 range library (such as `std::ranges::size()`) are functor objects (also called "customization point objects") which avoid interference with ADL.

To illustrate why I generally dislike ADL-discoverable helper functions, take the example of [`std::begin()`](https://en.cppreference.com/w/cpp/iterator/begin)/[`std::end()`](https://en.cppreference.com/w/cpp/iterator/end). As far as I understand, the *raison d'être* for these functions is that you want to be able to write generic looping code which works with plain arrays as well as containers. We can't just write `c.begin()` and `c.end()` because plain arrays don't have member functions. But `std::begin()`/`std::end()` conveniently calls the member function if one exists and also does the right thing for plain arrays, so we can use it like this:

```c++
template <typename C>
void someAlgorithm(C& container)
{
    for (auto i = std::begin(container), e = std::end(container); i != e; ++i) { ... }
}
void someCaller(void)
{
    int a[3]{ 1, 2, 3 };
    someAlgorithm(a);
}
```

Now, the [range-based for loop](https://en.cppreference.com/w/cpp/language/range-for) can also loop over range types which have no member functions `c.begin()` or `c.end()` but for which `begin()` and `end()` can be discovered per ADL. Let's change our algorithm accordingly, using the "two-step dance" to permit ADL:

```c++
template <typename C>
void someAlgorithm_ADL(C& container)
{
    using std::begin;
    using std::end;
    for (auto i = begin(container), e = end(container); i != e; ++i) { ... }
}
```

In practice, the algorithm is always called with STL containers, so people become sloppy and omit the `using` statements, but they never notice because they only test with `std::vector<>`:

```c++
template <typename C>
void someAlgorithm_Sloppy(C& container)
{
    for (auto i = begin(container), e = end(container); i != e; ++i) { ... }
}
```

But then a library like GSL adds a container-like type like `span<>`. Of course `someAlgorithm_Sloppy()` will fail to compile for `span<>`; but rather than fixing their bugs, users issue requests like https://github.com/microsoft/GSL/issues/810 demanding that `span<>` be equipped with ADL-discoverable `begin()`/`end()` overloads as well.

Another case in point: C++20 adds [`std::ssize()`](https://en.cppreference.com/w/cpp/iterator/size) which is carefully specified to work for every container with a `c.size()` method, so one can always write
```c++
for (std::ptrdiff_t i = 0, n = std::ssize(c); i != n; ++i) { ... }
```
and have it work for every existing container as well as for arrays. But sloppy code may omit the qualification "because it works,"
```c++
for (std::ptrdiff_t i = 0, n = ssize(c); i != n; ++i) { ... }
```
and thus be unnecessarily incompatible with any containers in pre-C++20 codebases.

petamas:
OK, I think I get it now, at least regarding `as_nullable` and zGSL. I'll modify my PR accordingly.

On a related note: should we have a "zero-overhead" `not_null` in GSL-Lite? Either with a separate name (`not_null_alias`, `not_null_zo`, `z_not_null`), or behind a `gsl_CONFIG` switch (choose between class and alias). It might help with testing.

Regarding begin/end/ssize: I think I understand your PoV, but not necessarily agree; I'll have to do a bit of research before I can formulate a full answer.

petamas:
@mbeutel 
Houston, we have a problem - defining the method in the `detail` namespace, then `using` it in the `gsl` namespace does not prevent ADL: https://godbolt.org/z/6nv4nc

What should we do?

petamas:
I think I found a workaround: https://godbolt.org/z/dYs514

#include<memory>

```
namespace gsl
{
    template<class T>
    struct not_null {};

    namespace detail
    {
        namespace prevent_adl
        {
            template<class T>
            void as_nullable(const not_null<T>& nn) {}
        }
    }

    using namespace gsl::detail::prevent_adl;
}

int main()
{
    gsl::not_null<std::unique_ptr<int>> nn;
    gsl::as_nullable(nn);
}
```

I was reading through https://en.cppreference.com/w/cpp/language/adl, and I found this part:

> The set of declarations found by ordinary unqualified lookup and the set of declarations found in all elements of the associated set produced by ADL, are merged, with the following special rules
> 1. using-directives in the associated namespaces are ignored

This seemed to indicate that your way should prevent ADL. However, after further inspection, I found out that only `using namespace XXX` counts as a "using-directive", while `using gsl::detail::as_nullable` is a "using-declaration". (Not to be confused with `using my_int = int`, which is an "alias declaration". I love this language! :D)

So, once again, I have a way of doing the fix.

mbeutel:
> On a related note: should we have a "zero-overhead" `not_null` in GSL-Lite?

I don't think we need that. *gsl-lite* already can be made almost zero-overhead with the contract violation config switches.

> Houston, we have a problem - defining the method in the detail namespace, then using it in the gsl namespace does not prevent ADL

Ouch. Right. `using namespace` is the way to go. (Or function objects; but without C++17 inline variables they raise lifetime issues we'd rather avoid.)

petamas:
(@mbeutel, I do not want to rush you, but I don't know if you noticed that I already added the ADL-related fix, so I'm leaving this comment.)

mbeutel:
Thank you. I get email notifications for any of your actions on this repo.

include/gsl/gsl-lite.hpp
```diff
@@ -2220,6 +2226,72 @@ not_null<U> make_not_null( not_null<U> const & u )
...
+{
+    typedef T nullable_type;
+
+    static nullable_type const & call(not_null<nullable_type> const & p)
+    {
+        return p.data_.ptr_;
+    }
+
+#if gsl_HAVE( RVALUE_REFERENCE )
+    static nullable_type && call(not_null<nullable_type> && p)
```
mbeutel, line 2263: This function should return `nullable_type`, not `nullable_type &&`, otherwise `as_nullable(std::move(nnp))` can be abused to get direct access to the underlying pointer.

kramermt:
@petamas : I'd love to have this on `master`, anything I can do to help get this finalized?  I;d be happy to take a pass at addressing @mbeutel 's feedback and submit a patch here, or make a PR to your fork.

petamas:
@kramermt: I was a bit preoccupied, but I'll get back to this PR in a few days. If I'm not done by Sunday, feel free to fix it up, but I'll try to be the first. :)

mbeutel:
Sorry for taking so long. I'll now take this PR as-is and apply my nit-picky changes through a separate PR. Thank you for the contribution.

Issue #252, opened 2020-07-27
GSL-Lite does not work on Apple Clang in Objective-C++ mode when exceptions are turned off


petamas:
The solution for #216 contains a minor bug. Currently, GSL-Lite uses either  `__EXCEPTIONS` or `__cpp_exceptions` to determine if exceptions are enabled during Clang compilation, depending on version:

```
#if gsl_BETWEEN(gsl_COMPILER_GNUC_VERSION, 1, 500) || gsl_BETWEEN(gsl_COMPILER_CLANG_VERSION, 1, 360) || gsl_COMPILER_APPLECLANG_VERSION
# ifdef __EXCEPTIONS
#  define gsl_HAVE_EXCEPTIONS  1
# else
#  define gsl_HAVE_EXCEPTIONS  0
# endif // __EXCEPTIONS
#elif gsl_COMPILER_GNUC_VERSION >= 500 || gsl_COMPILER_CLANG_VERSION >= 500
# ifdef __cpp_exceptions
#  define gsl_HAVE_EXCEPTIONS  1
# else
#  define gsl_HAVE_EXCEPTIONS  0
# endif // __cpp_exceptions
#elif gsl_COMPILER_MSVC_VERSION
# ifdef _CPPUNWIND
#  define gsl_HAVE_EXCEPTIONS  1
# else
#  define gsl_HAVE_EXCEPTIONS  0
# endif // _CPPUNWIND
#else
// For all other compilers, assume exceptions are always enabled.
# define  gsl_HAVE_EXCEPTIONS  1
#endif
```

This is broken in multiple ways:
- Clang [3.6, 5.0) is not handled at all (see `gsl_COMPILER_CLANG_VERSION >= 500`)
- Starting from Clang 3.6, `__EXCEPTIONS` is defined if either C++ or Objective-C exceptions are enabled,  which we handle in an OK way with vanilla Clang (apart from the previous bullet point), because we use `__cpp_exceptions` instead, but AppleClang is bundled with the pre-3.6 case, regardless of compiler version.

This means that GSL-Lite will not build if all of the following are true:
- AppleClang is used
- AppleClang is based on a Clang 3.6+ version
- C++ exceptions are disabled
- Objective-C exceptions are enabled
In this case, we derive that we _have_ C++ exception support, but in fact we do not, which causes all kinds of funny build errors:
```
.../gsl-lite.hpp:1229:5: error: cannot use 'throw' with exceptions disabled
throw fail_fast( message );
^
[...]

PR #255, opened 2020-07-27
Fix issue #252: build error on AppleClang10 when C++ exceptions are turned off

petamas:
Fix issue #252

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -292,18 +292,26 @@
 
 // Presence of language & library features:
 
-#if gsl_BETWEEN(gsl_COMPILER_GNUC_VERSION, 1, 500) || gsl_BETWEEN(gsl_COMPILER_CLANG_VERSION, 1, 360) || gsl_COMPILER_APPLECLANG_VERSION
-# ifdef __EXCEPTIONS
+#if gsl_COMPILER_CLANG_VERSION || gsl_COMPILER_APPLECLANG_VERSION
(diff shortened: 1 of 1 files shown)


include/gsl/gsl-lite.hpp
```diff
@@ -293,11 +293,28 @@
...
+#   define gsl_HAVE_EXCEPTIONS  1
+#  else
+#   define gsl_HAVE_EXCEPTIONS  0
+#  endif // __EXCEPTIONS && __has_feature(cxx_exceptions)
 # else
-#  define gsl_HAVE_EXCEPTIONS  0
-# endif // __EXCEPTIONS && __has_feature(cxx_exceptions)
+   // clang-cl doesn't define __EXCEPTIONS for MSVC compatibility (see https://reviews.llvm.org/D4065)
+   // Neither does clang in MS-compatiblity mode
+   // Let's hope no one tries to build Objective-C++ code using MS-compatibility mode or clang-cl
```
mbeutel, line 311: I haven't heard hear from [Project Islandwood](https://github.com/Microsoft/WinObjC/releases) in quite a while so I guess we shouldn't worry :)

petamas: I should've known there is something like that out there. :D We'll cross that bridge when we get there. I could come up with something that only fails if you use a pre-3.5 clang-cl building ObjectiveC++ with ObjC exceptions enabled, but C++ exceptions disabled, but I don't think it's worth it.

mbeutel: Neither do I.

mbeutel:
Darn, hit Return too fast and messed up the commit message.

Issue #254, opened 2020-07-27
gsl_CONFIG_DEFAULTS_VERSION=1 does not compile without exceptions 


petamas:
1. `gsl::narrow` is defined only if we have exception support, or `gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION` is set to 0. (See discussion in issue #52)
2. `gsl_CONFIG_DEFAULTS_VERSION=1` turns on `gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION`

This means `gsl::narrow` will not be defined in this case. However, the `gsl_lite` namespace unconditionally imports `gsl::narrow`, resulting in a compile error.

My proposed solution is to give a name to `gsl_HAVE( EXCEPTIONS ) || !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION` and use that to conditionally compile both `gsl::narrow`'s and `gsl_lite::narrow`'s definitions, but I do not know which prefix I should use - is it a gsl_CONFIG, a gsl_FEATURE, or something else?

I think @martinmoene migth be able to help me with this.

mbeutel:
> This means `gsl::narrow` will not be defined in this case. However, the `gsl_lite` namespace unconditionally imports `gsl::narrow`, resulting in a compile error.

Thank you for bringing up this oversight.


> My proposed solution is to give a name to `gsl_HAVE( EXCEPTIONS ) || !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION` and use that to conditionally compile both `gsl::narrow`'s and `gsl_lite::narrow`'s definitions, but I do not know which prefix I should use - is it a gsl_CONFIG, a gsl_FEATURE, or something else?

If I understand you correctly, you want

```c++
#define gsl_NOT_YET_NAMED gsl_HAVE( EXCEPTIONS ) || !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION
```

I think that would be neither a `gsl_CONFIG` nor a `gsl_FEATURE`; both are meant to be controllable by the user, which this macro wouldn't be. If anything, it could be named `gsl_HAVE_*`.

However, I'd rather avoid introducing a (non-internal, documented) name for it unless there is a compelling argument for exposing this mode to the library user. Would this be useful anwhere other than in the definition of `namespace gsl_lite`?

petamas:
Yes, it would only be used in two places:
- around the defintion of `gsl::narrow`
[...]

PR #262, opened 2020-07-29
Fix issue #254: gsl_CONFIG_DEFAULTS_VERSION=1 does not compile without exceptions

petamas:
Fix issue #254.

Also, replaced some direct gsl_CONFIG_XXX instances with gsl_CONFIG( XXX )

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -1631,18 +1631,21 @@ namespace detail {
 
 #endif
 
-#if gsl_HAVE( EXCEPTIONS ) || !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION
+// If narrow must throw (gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION), but exceptions are disabled (!gsl_HAVE_EXCEPTIONS), we cannot do anything
+#define gsl_NARROW_CAN_BE_DEFINED_ (gsl_HAVE( EXCEPTIONS ) || ! gsl_CONFIG( NARROW_THROWS_ON_TRUNCATION ))
+
+#if gsl_NARROW_CAN_BE_DEFINED_
 template< class T, class U >
-# if !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION && !defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
+# if !gsl_CONFIG( NARROW_THROWS_ON_TRUNCATION ) && !defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
 gsl_api
-# endif // !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION && !defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
+# endif // !gsl_CONFIG( NARROW_THROWS_ON_TRUNCATION ) && !defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
 inline T narrow( U u )
 {
     T t = static_cast<T>( u );
 
     if ( static_cast<U>( t ) != u )
     {
-# if gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION || defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
+# if gsl_CONFIG( NARROW_THROWS_ON_TRUNCATION ) || defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
         throw narrowing_error();
 # else
         std::terminate();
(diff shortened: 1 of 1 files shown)


include/gsl/gsl-lite.hpp
```diff
@@ -1631,18 +1631,21 @@ namespace detail {
...
 #endif
 
-#if gsl_HAVE( EXCEPTIONS ) || !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION
+// If narrow must throw (gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION), but exceptions are disabled (!gsl_HAVE_EXCEPTIONS), we cannot do anything
+#define gsl_NARROW_CAN_BE_DEFINED_ (gsl_HAVE( EXCEPTIONS ) || ! gsl_CONFIG( NARROW_THROWS_ON_TRUNCATION ))
+
+#if gsl_NARROW_CAN_BE_DEFINED_
 template< class T, class U >
-# if !gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION && !defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
+# if !gsl_CONFIG( NARROW_THROWS_ON_TRUNCATION ) && !defined( gsl_CONFIG_CONTRACT_VIOLATION_THROWS )
```
mbeutel, line 1639: I tend to avoid `gsl_HAVE()`, `gsl_CONFIG()` etc. in new code because I feel they don't add anything, and they are an impediment to searchability. It happened to me several times that I renamed a config switch (say, `gsl_CONFIG_A` to `gsl_CONFIG_B`) but failed to find some occurrences of `gsl_CONFIG( A )`. Of course the mistake goes unnoticed because the preprocessor implicitly assumes that `gsl_CONFIG_A`, which `gsl_CONFIG( A )` expands to, is 0 if there's no definition for it. Renaming is much more reliable if all references to `gsl_CONFIG_A` are spelled exactly `gsl_CONFIG_A` and can be found by global symbol search.

petamas: Well, I'm a sucker for consistency, so I think we should either remove these macro functions completely, or keep using them, but mixing the two appropaches does not seem wise.

However, you brought up this: 

> Of course the mistake goes unnoticed because the preprocessor implicitly assumes that gsl_CONFIG_A, which gsl_CONFIG( A ) expands to, is 0 if there's no definition for it.

I have once read an article about a workaround to this problem: by using function macros, you can disable the "default to 0" feature.

If for every config macro, we do something like this (note the `_F()`):

```
#ifdef gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION
# define gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION_F()  (gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION)
#else
# define gsl_CONFIG_NARROW_THROWS_ON_TRUNCATION_F()  (gsl_CONFIG_DEFAULTS_VERSION >= 1)
#endif
```

And we have a `gsl_CONFIG` like this: `#define gsl_CONFIG(feature)  ( gsl_CONFIG_##feature##_F() )`

Then using `gsl_CONFIG` detects misspellings, instead of hindering the rename process.

What do you think, should we pursue this approach?

petamas: Found the article: https://www.fluentcpp.com/2019/05/28/better-macros-better-flags/

mbeutel: I've recently come across a similar article:
https://thephd.github.io/versioning-and-other-boilerplate

Yes, I'd like to implement this. Here's a draft that catches wrong config values as well as typos:
https://gcc.godbolt.org/z/8PPhaM
What do you think?

petamas: Well, it looks interesting, but to be frank, I don't really like the error message. I mean, it can be reverse-engineered, but `Function-like macro 'gsl_CONFIG_XXX_F' is not defined` is easier to interpret. :) Although it can check for invalid features, which I like.

Would you be okay with spinning this conversation off to a new issue, and merging this PR as-is? Or should I unmake my `gsl_CONFIG_XXX`->`gsl_CONFIG(XXX)` changes?

mbeutel: Yes, let's take this to a separate issue. PR is fine.

martinmoene: > 
> 
> I tend to avoid `gsl_HAVE()`, `gsl_CONFIG()` etc. in new code because I feel they don't add anything, and they are an impediment to searchability. [...]

@mbeutel @petamas "It looked like a good idea at the time" :)

PR #263, opened 2020-07-29
Fix #258: not_null<T> should not have conversion to bool

### LINKED ISSUE #258: not_null<T> should not have conversion to bool
It's idiomatic to check a (raw or smart) pointer for `nullptr` by converting it to `bool`: `if (p) { ... }`. The purpose of `not_null` is to eliminate these checks when they're unnecessary because we already know the pointer cannot be `nullptr`, thus the conversion to `bool` will result in `true`.

However, the current version of the library allows `not_null<T>` to be converted to anything `T` can be converted into, including `bool`. This means that the following code happily compiles:

```
not_null<Thing*> p = ...;
if (p)
{
  // use p
} else {
  // !!! DEAD CODE !!!
}
```

I believe this goes directly against the purpose of this template. I propose to disallow conversion to `bool`, eg. by modifying `it_not_null_oracle` to `it_not_null_or_bool_oracle`. (The name is a work in progress.)

My main question is: should this behaviour be governed by a configuration macro, or should this be a breaking change? If the former, could you propose a good name for it? (I'm really bad at naming config switches.)

petamas:
Fix issue #258

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -1782,9 +1782,12 @@ struct element_type_helper< T* >
 #endif
 
 template< class T >
-struct is_not_null_oracle : std11::false_type { };
+struct is_not_null_or_bool_oracle : std11::false_type { };
 template< class T >
-struct is_not_null_oracle< not_null<T> > : std11::true_type { };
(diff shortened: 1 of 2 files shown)


mbeutel:
Good catch w.r.t. `is_assignable<>`.

PR #274, opened 2020-08-14
Add `require_not_null()`

### LINKED ISSUE #257: not_null<T> construction from T: is the "explicit" constructor explicit enough?
# Problem statement

Last year, I described Dropbox nn's constructor this way (see my comment in issue #200):

> - it does not have an nn(T) constructor
> - it does have an i_promise_i_checked_for_null_t tag type
> - it does have an nn(i_promise_i_checked_for_null_t, T) constructor
> - you're urged to create some macro named NN_CHECK (or any name you want to use, like CHECK_NOT_NULL), which checks for null, does whatever error handling you want to if the pointer is null, then calls the tagged constructor
> 
> This way, you must be pretty explicit about creating nn objects, but I think it's not much safer to use that version than ours (only advantage is that you can't construct an nn "by accident", using static_casts and such).

Since then, I have revised my opinion: I think it would be better if you couldn't create a `not_null` from a nullable without explicitly calling `make_not_null`. The following is the description of the evolution of my company's codebase, highlighting the problem with our current (explicit) constructor, then my proposed solution.

# Case study

## Original code

We had several old, pre-C++11 factory methods:
```
gsl::owner<Thing1*> createThing1(); // never returns NULL
gsl::owner<Thing2*> createThing2(); // never returns NULL
gsl::owner<Thing3*> createThing3(); // my return NULL based on configuration
```
The client code used `std::auto_ptr` to manage the memory:

```
std::auto_ptr<Thing1> thing1(createThing1());
std::auto_ptr<Thing2> thing2(createThing2());
std::auto_ptr<Thing3> thing3(createThing3());
```

## Introducing unique_ptr

After migrating to C++11, we changed our `auto_ptr`s to `unique_ptr`s, an made our factory methods return `unique_ptr`:

```
std::unique_ptr<Thing1> createThing1(); // never returns nullptr
std::unique_ptr<Thing2> createThing2(); // never returns nullptr
std::unique_ptr<Thing3> createThing3(); // my return nullptr based on configuration
// ...
std::unique_ptr<Thing1> thing1(createThing1());
std::unique_ptr<Thing2> thing2(createThing2());
std::unique_ptr<Thing3> thing3(createThing3());
```

Note that we did the `auto_ptr`->`unique_ptr` rather mindlessly, not changing the syntax in the client apart from changing the smart pointer's name.

## Introducing not_null

When `not_null` became able to handle `unique_ptr`, we changed the return type of the factory functions to `nn_unique_ptr<T>` (i.e. `gsl::not_null<std::unique_ptr<T>>`):

```
nn_unique_ptr<Thing1> createThing1(); // never returns nullptr
nn_unique_ptr<Thing2> createThing2(); // never returns nullptr
std::unique_ptr<Thing3> createThing3(); // my return nullptr based on configuration
// ...
nn_unique_ptr<Thing1> thing1(createThing1());
nn_unique_ptr<Thing2> thing2(createThing2());
nn_unique_ptr<Thing3> thing3(createThing3()); // !!! oooops !!!
```

During this transition, we made a mistake, and changed the type of `thing3` even though it should have been left as a plain old `unique_ptr`. The above code compiled without errors, and the mistake was only discovered during testing, when  `createThing3()` was configured to return `nullptr`.

# Proposed solution

The events described above led me to believe that `not_null<T>` should not have a converting constructor from T or any other nullable types; the conversion should always be done via explicitly calling `make_not_null()`. In that case, `nn_unique_ptr<Thing3> thing3(createThing3());` would result in a compile-time error instead of a run-time error.

This could be achieved by stealing the idea from Dropbox nn, and giving it our own spin:
- make the `T`->`not_null<T>` constructor private, and give it a tag argument
- make `make_not_null()` a friend of `not_null`

This way, only `make_not_null()` could be used to create `not_null`s, while not impacting the converting constructors between different `not_null` instantiations. Coupled with my proposal in issue #253, it would mean the only transition between nullable- and not_null-land would be through the two named functions `make_not_null` and `as_nullable`, making everything explicit and (mostly) mistake-free.

mbeutel:
Fixes #257.


mbeutel:
Closed for reasons discussed in #257. The CMake-related changes of this PR were migrated to #285.

PR #279, opened 2020-11-04
port to Keil(ARM Compiler v5)

woodsking2:
port it to Keil(ARM Compiler v5)
check it with --cpp11
armcc(ARM Compiler v5) has partially C++11 features supported.
The stl is only for c++ 03.
So it can't pass the compile with the newest gsl-lite.
I modified some code to let it work.

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -499,25 +499,53 @@
 #define gsl_CPP11_140_CPP0X_90   (gsl_CPP11_140 || (gsl_COMPILER_MSVC_VER >= 1500 && gsl_HAS_CPP0X))
 #define gsl_CPP11_140_CPP0X_100  (gsl_CPP11_140 || (gsl_COMPILER_MSVC_VER >= 1600 && gsl_HAS_CPP0X))
 
+#if defined(__ARMCC_VERSION) && __ARMCC_VERSION < 6000000
+#define gsl_HAVE_ARRAY                              0
+#define gsl_HAVE_INITIALIZER_LIST                   0
+#define gsl_HAVE_TYPE_TRAITS                        0
+#define gsl_HAVE_TR1_TYPE_TRAITS                    0
+#define gsl_HAVE_UNIQUE_PTR                         0
+#define gsl_HAVE_SHARED_PTR                         0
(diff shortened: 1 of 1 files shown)


mbeutel:
Thank you for your contribution.

I'm open towards adding support for more compilers. However, for proper support, we should introduce compiler detection macros as in https://github.com/gsl-lite/gsl-lite/blob/master/include/gsl/gsl-lite.hpp#L350..L397.

We would also need to make some changes to properly support the scenario where language and library version differently. I think the best approach would be to introduce a set of `gsl_STLxx_OR_GREATER` macros alongside the existing `gsl_CPPxx_OR_GREATER` macros, and to change our library feature detection macros accordingly. I can do some groundwork for this.

I understand that the ARM compiler is a commercial product, so we probably won't be able to add it to our CI test suite. This is unfortunate as inadvertent breakage will not be noticed in testing.

woodsking2:
> Thank you for your contribution.
> 
> I'm open towards adding support for more compilers. However, for proper support, we should introduce compiler detection macros as in https://github.com/gsl-lite/gsl-lite/blob/master/include/gsl/gsl-lite.hpp#L350..L397.
> 
> We would also need to make some changes to properly support the scenario where language and library version differently. I think the best approach would be to introduce a set of `gsl_STLxx_OR_GREATER` macros alongside the existing `gsl_CPPxx_OR_GREATER` macros, and to change our library feature detection macros accordingly. I can do some groundwork for this.
> 
> I understand that the ARM compiler is a commercial product, so we probably won't be able to add it to our CI test suite. This is unfortunate as inadvertent breakage will not be noticed in testing.

Thanks for reply.
We can use 

> #if defined(__ARMCC_VERSION) && __ARMCC_VERSION < 6000000


to detect it.
I will update it introduce compiler detection macros as in https://github.com/gsl-lite/gsl-lite/blob/master/include/gsl/gsl-lite.hpp#L350..L397.

mbeutel:
Sorry for lack of progress. I plan to get back to it before next week.

mbeutel:
@woodsking2: could you please test whether #293 works for you with ARMCC v5? If so, #293 would supersede this PR.

woodsking2:
Close. 
supersede  by #293

PR #294, opened 2021-03-23
Add `gsl_Assert()`, `gsl_AssertAudit()`, `gsl_FailFast()`, contract violation modes `ASSERTS` and `TRAPS`

mbeutel:
This PR makes the following changes:

- Adds additional invariant checking macros `gsl_Assert()`, `gsl_AssertAudit()` to express checks which are neither pre- nor postconditions.

- Adds fail-fast operation `gsl_FailFast()` which is guaranteed to terminate normal execution in some way (its behavior can be configured, but unlike the contract checks it cannot be suppressed entirely), enabling uses like
    ```c++
    enum class Color { red, green, blue };
    std::string_view to_string( Color color )
    {
        switch ( color )
        {
        case Color::red: return "red";
        case Color::green: return "green";
        case Color::blue: return "blue";
        }
        gsl_FailFast();  // guaranteed to throw or terminate ⇒ no compiler warning about function not returning a value
    }
    ```

- Adds contract violation handling configuration option `gsl_CONFIG_CONTRACT_VIOLATION_TRAPS` which makes `gsl_Expects()`/`gsl_Ensures()`/`gsl_Assert()` (and the `Audit` variants if audit mode is enabled) and `gsl_FailFast()` execute a trap instruction (`__fastfail()` for MSVC, `__builtin_trap()` for GCC and Clang) in case of a contract violation.

- Adds contract violation handling configuration option `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` which implements `gsl_Expects()`/`gsl_Ensures()`/`gsl_Assert()` (and the `Audit` variants if audit mode is enabled) in terms of the [`assert()`](https://en.cppreference.com/w/cpp/error/assert) macro from the standard library. This has the benefits that *all* assertions can be globally suppressed with a single macro (i.e. `NDEBUG`), and that `assert()` typically prints an informative error message which contains the contract check expression and not just a dull generic message like `std::terminate()`.


mbeutel:
Although I'd very much prefer to make `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` the version-1 default (of course `gsl_CONFIG_CONTRACT_VIOLATION_TERMINATES` would remain the version-0 default), this would render our default behavior slightly non-compliant because the Core Guidelines specify that contract violation leads to termination unless configured otherwise, cf. https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslassert-assertions. Therefore I'm keeping this change for another PR.

PR #297, opened 2021-03-26
Added std::hash<gsl::not_null<...>>

### LINKED ISSUE #296: `not_null<>` should be hashable
...if the pointer type is hashable, that is (cf. https://github.com/microsoft/GSL/issues/914).

mbs-c:
Closes #296

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -704,11 +704,11 @@
 #endif
 
 #if !gsl_HAVE( NOEXCEPT ) || defined( gsl_TESTING_ )
-# define gsl_noexcept           /*noexcept*/
-# define gsl_noexcept_iff(expr) /*noexcept(noexcept(expr))*/
+# define gsl_noexcept          /*noexcept*/
+# define gsl_noexcept_if(expr) /*noexcept(expr)*/
 #else
 # define gsl_noexcept           noexcept
-# define gsl_noexcept_iff(expr) noexcept(noexcept(expr))
+# define gsl_noexcept_if(expr) noexcept(noexcept(expr))
 #endif
 
 #if gsl_HAVE( NULLPTR )
@@ -4572,7 +4572,7 @@ struct hash< ::gsl::not_null< T > > : public ::gsl::detail::conditionally_enable
 {
 public:
     gsl_NODISCARD gsl_constexpr std::size_t
-    operator()( const ::gsl::not_null< T >& v ) const gsl_noexcept_iff( hash<T>()( declval<T>() ) )
+    operator()( const ::gsl::not_null< T >& v ) const gsl_noexcept_if(gsl_noexcept( hash<T>()(declval<T>() ) ) )
     {
         return hash<T>()( ::gsl::as_nullable( v ) );
     }
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -704,11 +704,11 @@
 #endif
 
 #if !gsl_HAVE( NOEXCEPT ) || defined( gsl_TESTING_ )
-# define gsl_noexcept          /*noexcept*/
-# define gsl_noexcept_if(expr) /*noexcept(expr)*/
+# define gsl_noexcept            /*noexcept*/
+# define gsl_noexcept_if( expr ) /*noexcept( expr )*/
 #else
-# define gsl_noexcept           noexcept
-# define gsl_noexcept_if(expr) noexcept(noexcept(expr))
+# define gsl_noexcept            noexcept
+# define gsl_noexcept_if( expr ) noexcept( expr )
 #endif
 
 #if gsl_HAVE( NULLPTR )
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -704,9 +704,11 @@
 #endif
 
 #if !gsl_HAVE( NOEXCEPT ) || defined( gsl_TESTING_ )
-# define gsl_noexcept /*noexcept*/
+# define gsl_noexcept           /*noexcept*/
+# define gsl_noexcept_iff(expr) /*noexcept(noexcept(expr))*/
 #else
-# define gsl_noexcept noexcept
+# define gsl_noexcept           noexcept
+# define gsl_noexcept_iff(expr) noexcept(noexcept(expr))
 #endif
 
 #if gsl_HAVE( NULLPTR )
@@ -4529,7 +4531,52 @@ typedef basic_zstring_span< wchar_t const > cwzstring_span;
 
 #if gsl_HAVE( HASH )
 
+//
+// std::hash specializations for GSL types
+//
+
+namespace gsl {
+
+namespace detail {
+
+//
+// Helper struct for std::hash specializations
+//
+
+template<bool Condition>
+struct conditionally_enabled_hash
+{
+
+};
+
+// disabled as described in [unord.hash]
+template<>
+struct conditionally_enabled_hash< false >
+{
+gsl_is_delete_access:
+    conditionally_enabled_hash() gsl_is_delete;
+    conditionally_enabled_hash( conditionally_enabled_hash const & ) gsl_is_delete;
+    conditionally_enabled_hash( conditionally_enabled_hash && ) gsl_is_delete;
+    conditionally_enabled_hash & operator=( conditionally_enabled_hash const & ) gsl_is_delete;
+    conditionally_enabled_hash & operator=( conditionally_enabled_hash && ) gsl_is_delete;
+};
+
+} // namespace detail
+
+} // namespace gsl
+
 namespace std {
+
+template<typename T>
+struct hash< ::gsl::not_null< T > > : public ::gsl::detail::conditionally_enabled_hash< is_default_constructible< hash< T > >::value >
+{
+public:
+    gsl_NODISCARD gsl_constexpr std::size_t
+    operator()( const ::gsl::not_null< T >& v ) const gsl_noexcept_iff( hash<T>()( declval<T>() ) )
+    {
+        return hash<T>()( ::gsl::as_nullable( v ) );
+    }
+};
 
 template<>
 struct hash< ::gsl::byte >
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -382,7 +382,7 @@
 // AppleClang 10.0.1  __apple_build_version__ == 10010046  gsl_COMPILER_APPLECLANG_VERSION == 1001  (Xcode 10.2, 10.2.1, 10.3)       (LLVM  7.0.0)
 // AppleClang 11.0.0  __apple_build_version__ == 11000033  gsl_COMPILER_APPLECLANG_VERSION == 1100  (Xcode 11.1, 11.2, 11.3, 11.3.1) (LLVM  8.0.0)
 // AppleClang 11.0.3  __apple_build_version__ == 11030032  gsl_COMPILER_APPLECLANG_VERSION == 1103  (Xcode 11.4, 11.4.1, 11.5, 11.6) (LLVM  9.0.0)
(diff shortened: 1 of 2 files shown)


mbs-c:
The VS 2015 build is easily fixed by replacing `#if gsl_CPP11_140` with `#if ! gsl_BETWEEN( gsl_COMPILER_MSVC_VERSION, 1, 141 )`. But AppleClang doesn't like it in any version, so maybe I need to write a better test that doesn't assume the existence of a disabled specialization.

> For every type Key for which neither the library nor the user provides an enabled specialization std::hash<Key>, that specialization exists and is disabled.

Is https://en.cppreference.com/w/cpp/utility/hash lying?

mbs-c:
Indeed, this only seems to be true since C++17:

 * [\[unord.hash\], C++11](https://timsong-cpp.github.io/cppwp/n3337/unord.hash)
 * [\[unord.hash\], C++14](https://timsong-cpp.github.io/cppwp/n4140/unord.hash)
 * [\[unord.hash\], C++17](https://timsong-cpp.github.io/cppwp/n4659/unord.hash)

This specific was introduced in [P0513R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0513r0.pdf).

I guess that means a more complicated mechanism is needed to conditionally disable `std::hash<not_null<>>`.

test/not_null.t.cpp
```diff
@@ -1486,23 +1486,23 @@ CASE( "not_null<>: Able to deduce element_type of user-defined smart pointers ev
...
-    EXPECT(( std::is_default_constructible< std::hash< gsl::not_null< std::unique_ptr< int > > > >::value ));
-    EXPECT(( std::is_default_constructible< std::hash< gsl::not_null< std::shared_ptr< int > > > >::value ));
+    EXPECT(( std::is_default_constructible< std::hash< not_null< int* > > >::value ));
+    EXPECT(( std::is_default_constructible< std::hash< not_null< std::unique_ptr< int > > > >::value ));
+    EXPECT(( std::is_default_constructible< std::hash< not_null< std::shared_ptr< int > > > >::value ));
 #if gsl_CPP11_140
     // doesn't even compile in VS 2013, but at least the error message is somewhat descriptive:
     //     error C2338: The C++ Standard doesn't provide a hash for this type.
-    EXPECT_NOT(( std::is_default_constructible< std::hash< gsl::not_null< NormalPtr< int > > > >::value ));
+    EXPECT_NOT(( std::is_default_constructible< std::hash< not_null< NormalPtr< int > > > >::value ));
```
mbeutel, line 1505: Due to the C++17 wording changes you cited, I think the `EXPECTS_NOT()` check here should be guarded with `#if gsl_STDLIB_CPP17_OR_GREATER`. As far as I can see, the AppleClang tests succeed with `-std=c++17` but fail with e.g. `-std=c++11`.

mbeutel:
Thanks for the contribution. Looks very good overall.

> I guess that means a more complicated mechanism is needed to conditionally disable `std::hash<not_null<>>`.

I don't think so. Does the standard state any requirements for `std::hash<>` specializations prior to C++17? As far as I can see it does not, so having C++17 behavior even for older standards is not a problem. Most standard library implementations except for AppleClang's libc++ seem to retain the C++17 behavior even when compiling as C++11 or C++14, otherwise your `EXPECT_NOT()` test would have broken them.

include/gsl/gsl-lite.hpp
```diff
@@ -704,9 +704,11 @@
 #endif
 
 #if !gsl_HAVE( NOEXCEPT ) || defined( gsl_TESTING_ )
-# define gsl_noexcept /*noexcept*/
+# define gsl_noexcept           /*noexcept*/
+# define gsl_noexcept_iff(expr) /*noexcept(noexcept(expr))*/
```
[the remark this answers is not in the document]
mbs-c: The idea behind the naming was that the function would be `noexcept` if and only if the expression was `noexcept` as well, but I'm not particularly happy with the name, either. I wanted to avoid having nested macros `gsl_noexcept_if( gsl_noexcept( ... ) )`, but if you prefer that, I'll write it that way.

mbs-c:
> Does the standard state any requirements for `std::hash<>` specializations prior to C++17?

Unless I misread the test results, the major standard libraries already seem to exhibit the "correct" behavior in C++14 mode, probably as a result of the enum support in `std::hash<>`. Therefore, it's not entirely clear to me whether this behavior should be enabled from C++14 or C++17, which is why my gut feeling would be that I should make it work with all language versions to avoid confusion/problems.

include/gsl/gsl-lite.hpp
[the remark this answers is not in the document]
mbeutel: A function/macro named "*_if" or "*_iff" that *doesn't* expect a Boolean argument feels weird to me.

For a Boolean argument "A", the "if and only if" is redundant: we are *defining* a Boolean property "B" of the function (stating whether the function is `noexcept`). It is true that "B" can be defined implicitly by saying "(A ⇒ B) ∧ (¬A ⇒ ¬B)", but why would we if we can simply say "B := A"?

test/not_null.t.cpp
mbs-c: I just pushed a test commit with `#if gsl_STDLIB_CPP14_OR_GREATER && ! gsl_BETWEEN(gsl_COMPILER_MSVC_VERSION, 1, 141)`. Let's see whether that works on all tested platforms.

(I'll squash the test commits when the PR is ready to merge.)

mbeutel: (No need to squash, I'll do that upon merging.)

include/gsl/gsl-lite.hpp
```diff
@@ -704,11 +704,11 @@
...
 
 #if !gsl_HAVE( NOEXCEPT ) || defined( gsl_TESTING_ )
-# define gsl_noexcept           /*noexcept*/
-# define gsl_noexcept_iff(expr) /*noexcept(noexcept(expr))*/
+# define gsl_noexcept          /*noexcept*/
+# define gsl_noexcept_if(expr) /*noexcept(expr)*/
 #else
 # define gsl_noexcept           noexcept
-# define gsl_noexcept_iff(expr) noexcept(noexcept(expr))
+# define gsl_noexcept_if(expr) noexcept(noexcept(expr))
```
mbeutel, line 711: Your definition here still has the second `noexcept()`...

include/gsl/gsl-lite.hpp
```diff
@@ -4572,7 +4572,7 @@ struct hash< ::gsl::not_null< T > > : public ::gsl::detail::conditionally_enable
 {
 public:
     gsl_NODISCARD gsl_constexpr std::size_t
-    operator()( const ::gsl::not_null< T >& v ) const gsl_noexcept_iff( hash<T>()( declval<T>() ) )
+    operator()( const ::gsl::not_null< T >& v ) const gsl_noexcept_if(gsl_noexcept( hash<T>()(declval<T>() ) ) )
```
mbeutel, line 4575: ...so this statement is expanded to `noexcept(noexcept(noexcept(...)))`, which is always `true`.

mbeutel (commented):
(Also, a nitpick: please stick with the spacing conventions for macros:
```c++
#define macro( expr )  something( expr )
```
Spaces inside parentheses, double space between macro signature and substitution.)

include/gsl/gsl-lite.hpp
mbs-c: I guess I need a coffee, sorry...

test/not_null.t.cpp
mbs-c: So, it seems like VS 2015 is the only exception among the tested compilers. All others (including VS 2017 with `/std:c++14`) define `std::hash` as required by the C++17 specification.

mbs-c:
Regarding style: The macro was an oversight, but a `.clang-format` file would be nice. The spacing inside angle brackets seemed inconsistent in the existing code, so I wasn't quite sure how to format my change.

mbeutel:
> a `.clang-format` file would be nice.

Agreed. I always got stuck halfway when trying to put one together, but if someone with more experience could contribute a .clang-format file that reasonably approximates this style, I'd be grateful.

> The spacing inside angle brackets seemed inconsistent in the existing code, so I wasn't quite sure how to format my change.

The spacing style in this project doesn't match my personal preferences; I find `f( x, y )` borderline but `vector< pair< int, int > >` close to unreadable. The latter was already used inconsistently when I inherited the project, so I chose not to be too picky about spacing in `<>` template brackets.

martinmoene:
Regarding style: :D

@mbeutel Please feel free to change the coding style to your liking, what works best for you.

Different people have different needs to see the information in the 'wall of text' and the current spacing and alignment are still attuned to what supported my needs 'back then.' I am aware that people who are capable of fast(er) reading prefer a much more dense layout and may have a lesser need for alignment.

---

Up to now I've not been able to convince clang-format to produce an alignment somewhat like the following examples to accommodate for function declarations or calls to be spread over multiple lines that breath 'structural honesty' as Kevlin Henney calls it [1]:

```
template<
    typename T, typename U, typename V
>
void procedure(
    T t, U u, V v
)
{
    use(t, u, v);
}
```

and for a longer list of arguments:

```
template
<
    typename T
    , typename U
    , typename V
>
void procedure
(
    T t
    , U u
    , V v
)
{
    use
    (
        t
        , u
        , v
    );
}
```

[1] Kevlin Henney. Seven Ineffective Coding Habits of Many Programmers, '[Unsustainable spacing](https://vimeo.com/97329157#t=14m32s)', 2014. Vimeo.

Likely not the first time (nor presumably the last) I'm referring to 'Unsustainable spacing' from one of the more than seven versions of  'Seven Ineffective Coding Habits of Many Programmers' :)

mbeutel:
Martin, thank you for chiming in.

I guess my comment on the spacing style was more emotional than it needed to be; sorry about that.

Although I tend to obey slightly different spacing and indentation rules in my other projects, I am not opposed to *gsl-lite*'s style. The style is consistent and makes sense; the code feels like it was written by somebody who cared about the visual presentation. When extending the code, it seemed more natural to respect the unwritten rules expressed in the code rather than just imposing my own.

The points made by Kevlin Henney on indentation and alignment I totally agree with.

Regarding spacing, my preference is indeed to keep code a little more dense. But *gsl-lite*'s style doesn't bother me, and I even grew to like the spaces-in-parentheses rule, i.e. `if ( x != ( y == z ) )` rather than `if (x != (y == z))`. (I have to retract my "borderline" characterization above.)

The only part I dislike is abundance of spaces in angle brackets. In my opinion, using too many spaces in type names impedes the visual recognition of structure. Consider the basic structural elements in function and variable declarations:

```
TypeName variable = expression;
ReturnType functionName( ArgType argName );
```

The separation between `TypeName` and `variable` is usually established with a single space. This works fine even with unary templates:

```
std::vector<int> variable = expression;
gsl::not_null<std::unique_ptr<Resource>> functionName( std::vector<int> const& argName );
```

When using templates with more than one parameter, the rendering already becomes visually ambiguous: we need to visually "count angle brackets" to understand out that the second space is the type–name delimiter:

```
std::vector<std::pair<int, int>> variable = expression;
```

But with the space-inside-angle-brackets rule, spaces become so abundant that (for me) they lose their structural guidance function:

```
std::vector< int > variable = expression;
gsl::not_null< std::unique_ptr< Resource > > functionName( std::vector< int > const & argName );
```

It's true that we can somewhat make up for this by using line breaks and indentation,

```
gsl::not_null< std::unique_ptr< Resource > >
functionName(
    std::vector< int > const & argName );
```

But I think there is value in a style that doesn't depend on line breaks to make extremely simple declarations like `std::vector<int> variable` visually unambiguous.

martinmoene:
Thanks for your thoughtful remarks Moritz. 

I've come to like somewhat pointed remarks when they come 'straight from the heart' :), there's no need to apologize at all.

I agree with your observations and something along the line of 

```
std::vector<std::pair<int, int>> variable = functionName( first, second, third );  // Note 1
``` 

indeed seems to serve readability better than

```
std::vector< std::pair< int, int > > variable = functionName( first, second, third );
``` 

Note 1: And then there's C++98 that requires `...> >` :)

PR #301, opened 2021-03-31
Make `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` the version-1 default

mbeutel:
The contract violation handling configuration option `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` implements `gsl_Expects()`/`gsl_Ensures()`/`gsl_Assert()` (and the `Audit` variants if audit mode is enabled) in terms of the [`assert()`](https://en.cppreference.com/w/cpp/error/assert) macro from the standard library. This has the benefits that *all* assertions can be globally suppressed with a single macro (i.e. `NDEBUG`), and that `assert()` prints an informative error message which contains the contract check expression.

Because it is so convenient, I'd like to make `gsl_CONFIG_CONTRACT_VIOLATION_ASSERTS` the default in the next major version. However, as pointed out in #294, this would render our default behavior slightly non-compliant because the Core Guidelines specify that contract violation leads to termination unless configured otherwise, cf. https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslassert-assertions.


mbeutel:
CMake (and other build tools as well, I suppose) defines `NDEBUG` by default for the "Release", "RelWithDebInfo", and "MinSizeRel" build configurations, so implementing contract violation checks with `assert()` by default would silently remove runtime checks from release builds everywhere, which is not something I want to do. Therefore I won't make this change.

Issue #264, opened 2020-07-30
Add -fno-exceptions builds to CI


petamas:
There were multiple issues where GSL-Lite did not compile if exception support was turned off: #216, #252, #254 .

To prevent these kinds of issues, a new mode (apart from v0/v1) should be introduced to our CI pipeline.

See my proposal [here](https://github.com/gsl-lite/gsl-lite/issues/252#issuecomment-664428856):

> I think instead of having a separate test suite, we should adapt our already existing test suite to be able to be tested with or without exceptions:
> 
> * introduce a new macro called `EXPECT_FAIL` in `gsl_lite.t.hpp`
> * if exceptions are enabled, `EXPECT_FAIL` should be defined using `EXPECT_THROW`
> * if exceptions are disabled, `EXPECT_FAIL` should be defined using an implementation of `fail_fast_assert_handler ` (eg. `fail_fast_assert_handler` sets a global error flag, and `EXPECT_FAIL` contains an `EXPECT` on that)
> * replace `EXPECT_THROW` with `EXPECT_FAIL` in gsl_lite.t.hpp (where applicable, i.e. everywhere except test for `narrow`
> * modify the cmake files to build&run all tests both with and without exceptions, defining the corresponding `gsl_CONFIG_CONTRACT_VIOLATION_XXX` macro

mbeutel:
Of course I'm all for having CI coverage of the no-exception mode. However, a question:

What benefit would we get from modifying the test suite to run with exceptions disabled? The semantic behavior of *gsl-lite* doesn't depend on the contract violation handling mode, so given that we have working tests in throw-on-contract-violation mode, wouldn't it be enough if the tests *compiled* in (exception-less) fail-fast mode?

petamas:
I haven't checked, but I'm almost sure that building the tests is slower than running them, so in my opinion the real question is: why NOT run the tests in exceptionless mode?
Naturally, for my purposes, it's enough to build the tests, but I see no reason not to run them.

mbeutel:
> I haven't checked, but I'm almost sure that building the tests is slower than running them, so in my opinion the real question is: why NOT run the tests in exceptionless mode?

Because then we don't need to adapt the test suite to exception-less execution. No need for `EXPECT_FAIL()`, no need to work around tests that run into UB if precondition assertions don't abort on violation.

mbeutel:
[...]

PR #308, opened 2021-06-16
Basic NVHPC compiler support.

olupton:
`gsl-lite` does not work out of the box with the NVHPC compilers on my system:
```
"include/gsl/gsl-lite.hpp", line 1394: error: identifier "__ORDER_LITTLE_ENDIAN__" is undefined
      little = __ORDER_LITTLE_ENDIAN__,
               ^

"include/gsl/gsl-lite.hpp", line 1395: error: identifier "__ORDER_BIG_ENDIAN__" is undefined
      big    = __ORDER_BIG_ENDIAN__,
               ^

"include/gsl/gsl-lite.hpp", line 1396: error: identifier "__BYTE_ORDER__" is undefined
      native = __BYTE_ORDER__
               ^
```

This is a minimal patch to recognise the NVHPC compilers as distinct from GCC and assume little-endian order for them.

I tried to build and run the full set of `gsl-lite` tests and examples; there were no errors but a few warnings are repeated many times. For example:
```
"gsl-lite/test/not_null.t.cpp", line 694: warning: variable "<unnamed>::__lest_registrar__694" was declared but never referenced
  CASE( "not_null<>: Allows to construct a const pointer from a not_null related pointer type (shared_ptr)" )
  ^
```

### DIFF
--- a/test/assert.t.cpp
+++ b/test/assert.t.cpp
@@ -36,6 +36,9 @@ enum Color
 #endif
 { red, green, blue };
 
+#if gsl_COMPILER_NVHPC_VERSION
+# pragma diag_suppress 941 // Suppress:  missing return statement at end of non-void function "..."
+#endif
 std::string colorToString( Color color )
 {
     switch (color)
@@ -46,6 +49,10 @@ std::string colorToString( Color color )
     }
     gsl_FailFast();  // this should keep the compiler from issuing a warning about not returning a value
 }
+#if gsl_COMPILER_NVHPC_VERSION
+# pragma diag_default 941
+#endif
+
 
 struct ConvertibleToBool
 {
--- a/test/assert.t.cpp
+++ b/test/assert.t.cpp
@@ -36,7 +36,7 @@ enum Color
 #endif
 { red, green, blue };
 
-#if gsl_COMPILER_NVHPC_VERSION
+#if gsl_COMPILER_NVHPC_VERSION && !gsl_CPP11_OR_GREATER
 # pragma diag_suppress 941 // Suppress:  missing return statement at end of non-void function "..."
 #endif
 std::string colorToString( Color color )
@@ -49,7 +49,7 @@ std::string colorToString( Color color )
     }
     gsl_FailFast();  // this should keep the compiler from issuing a warning about not returning a value
 }
-#if gsl_COMPILER_NVHPC_VERSION
+#if gsl_COMPILER_NVHPC_VERSION && !gsl_CPP11_OR_GREATER
 # pragma diag_default 941
 #endif
 
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -478,7 +478,7 @@
 # define gsl_COMPILER_CLANG_VERSION       0
 #endif
 
-#if defined( __GNUC__ ) && ! defined( __clang__ )
+#if defined( __GNUC__ ) && ! defined( __clang__ ) && ! defined( __NVCOMPILER )
 # define gsl_COMPILER_GNUC_VERSION  gsl_COMPILER_VERSION( __GNUC__, __GNUC_MINOR__, __GNUC_PATCHLEVEL__ )
 #else
 # define gsl_COMPILER_GNUC_VERSION  0
@@ -488,6 +488,13 @@
 # define gsl_COMPILER_NVCC_VERSION  ( __CUDACC_VER_MAJOR__ * 10 + __CUDACC_VER_MINOR__ )
 #else
 # define gsl_COMPILER_NVCC_VERSION  0
+#endif
+
+// NVHPC 21.2  gsl_COMPILER_NVHPC_VERSION == 2120
+#if defined( __NVCOMPILER )
+# define gsl_COMPILER_NVHPC_VERSION  gsl_COMPILER_VERSION( __NVCOMPILER_MAJOR__, __NVCOMPILER_MINOR__, __NVCOMPILER_PATCHLEVEL__ )
+#else
+# define gsl_COMPILER_NVHPC_VERSION  0
 #endif
 
 #if defined( __ARMCC_VERSION )
@@ -1386,7 +1393,7 @@ struct identity
 # if gsl_HAVE( ENUM_CLASS )
 enum class endian
 {
-#  if defined( _WIN32 )
+#  if defined( _WIN32 ) || gsl_COMPILER_NVHPC_VERSION
     little = 0,
     big    = 1,
     native = little
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -1049,7 +1049,7 @@
 # include <intrin.h>
 #endif
 
-#if gsl_HAVE( ENUM_CLASS ) && gsl_COMPILER_ARMCC_VERSION
+#if gsl_HAVE( ENUM_CLASS ) && ( gsl_COMPILER_ARMCC_VERSION || gsl_COMPILER_NVHPC_VERSION )
 # include <endian.h>
 #endif
 
@@ -1393,7 +1393,7 @@ struct identity
 # if gsl_HAVE( ENUM_CLASS )
 enum class endian
 {
-#  if defined( _WIN32 ) || gsl_COMPILER_NVHPC_VERSION
+#  if defined( _WIN32 )
     little = 0,
     big    = 1,
     native = little
@@ -1401,7 +1401,7 @@ enum class endian
     little = __ORDER_LITTLE_ENDIAN__,
     big    = __ORDER_BIG_ENDIAN__,
     native = __BYTE_ORDER__
-#  elif gsl_COMPILER_ARMCC_VERSION
+#  elif gsl_COMPILER_ARMCC_VERSION || gsl_COMPILER_NVHPC_VERSION
     // from <endian.h> header file
     little = __LITTLE_ENDIAN,
     big    = __BIG_ENDIAN,
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -557,8 +557,8 @@
 #   define gsl_HAVE_EXCEPTIONS  0
 #  endif
 # endif
-#elif gsl_COMPILER_GNUC_VERSION
-# if gsl_BETWEEN(gsl_COMPILER_GNUC_VERSION, 1, 500)
+#elif defined( __GNUC__ )
+# if __GNUC__ < 5
 #  ifdef __EXCEPTIONS
 #   define gsl_HAVE_EXCEPTIONS  1
 #  else
@@ -570,7 +570,7 @@
 #  else
 #   define gsl_HAVE_EXCEPTIONS  0
 #  endif // __cpp_exceptions
-# endif // gsl_BETWEEN(gsl_COMPILER_GNUC_VERSION, 1, 500)
+# endif // __GNUC__ < 5
 #elif gsl_COMPILER_MSVC_VERSION
 # ifdef _CPPUNWIND
 #  define gsl_HAVE_EXCEPTIONS  1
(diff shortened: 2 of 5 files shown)


mbeutel:
Thank you for the contribution!

I didn't have any exposure to NVHPC yet; do you know which frontend (Clang, EDG) its C++ compiler it is based on? With that information, we might be able to get rid of the warnings. *lest* (our test framework) defines the following macro to suppress the maybe-unused diagnostic:

```c++
// Indicate argument as possibly unused, if possible:

#if __has_cpp_attribute(maybe_unused) && lest_CPP17_OR_GREATER
# define lest_MAYBE_UNUSED(ARG)  [[maybe_unused]] ARG
#elif defined (__GNUC__)
# define lest_MAYBE_UNUSED(ARG)  ARG __attribute((unused))
#else
# define lest_MAYBE_UNUSED(ARG)  ARG
#endif
```

(Of course we could also change *lest* to instead use the portable "cast-to-`void`" approach to suppress this warning, but I'd rather not invest too much. Besides, parts of *gsl-lite* might also benefit from frontend awareness.)

Does NVHPC come with its own standard library, or does it reuse the library that comes with another compiler (like ICC or NVCC do)? Does it know the \<endian.h\> header file? I'd be a little wary to hard-code little-endianness for this compiler given that it already supports several non-x86 targets (POWER, ARM) which I know nothing about, so I'd prefer to extract this information from the environment.

(The CI failure seems unrelated; I suspect that the most recent version of VS requires CUDA 11.3, so I need to update that.)

olupton:
The NVHPC compilers are an evolution of the older PGI compilers (https://www.pgroup.com/, https://en.wikipedia.org/wiki/PGI_compiler), I am not familiar with the details (the available options make me suspect it is somehow based on EDG).

I had a quick play around with `lest_MAYBE_UNUSED`, but did not manage to suppress the warning. I think one can in principle use a pair of `_Pragma("diag_suppress 177")` and `_Pragma("diag_default 177")`, but I did not manage to insert these into the definition of `CASE` in a way that actually worked. IIUC the error is about the instance of `lest::add_test` called `lest_REGISTRAR`, not the argument named `lest_env`:
```c++
#define lest_CASE( specification, proposition ) \
    static void lest_FUNCTION( lest::env & ); \
    namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); } \
    static void lest_FUNCTION( lest_MAYBE_UNUSED( lest::env & lest_env ) )
```

A brute force approach of disabling this warning for all tests revealed a couple more errors:
```
"gsl/gsl-lite.hpp", line 2244: warning #186-D: pointless comparison of unsigned integer with zero
      gsl_Expects( ( ::gsl::detail::is_same_signedness<T, U>::value || ( t < T() ) == ( u < U() ) ) );
      ^
          detected during instantiation of "T gsl::narrow_failfast<T,U>(U) [with T=size_t, U=std::ptrdiff_t]" at line 825 of "test/span.t.cpp"
```
and
```
"test/assert.t.cpp", line 48: warning #941-D: missing return statement at end of non-void function "<unnamed>::colorToString"
  }
  ^
```

NVHPC doesn't ship its own standard library; [it uses one from GCC](https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-user-guide/index.html#lib-std-cpp-template) on Linux. I assume this is also why with `gsl-lite` `master` then `gsl_COMPILER_GNUC_VERSION` was set even with NVHPC.

I switched it to use `endian.h`, I am not sure how portable this is -- on my system then this works (it picks up `/usr/include/endian.h`; NVHPC does not ship its own version).

For now I just pushed a brute-force change to disable the warnings. This only affects the tests.

mbeutel:
Thank you, very useful feedback.

Perhaps we could just change the definition in *lest* to
```diff
  #define lest_CASE( specification, proposition ) \
      static void lest_FUNCTION( lest::env & ); \
-     namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); } \
+     namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); (void) lest_REGISTRAR; } \
      static void lest_FUNCTION( lest_MAYBE_UNUSED( lest::env & lest_env ) )
```
to suppress the warning?

Regarding the signedness warning, I already had to introduce `#ifdef` branches to circumvent it for NVCC; look for calls to `::gsl::detail::have_same_sign()` in `narrow<>()` and `narrow_failfast<>()`. I suppose you could simply take these branches for the NVHPC compiler as well.

The warning about the missing return statement probably stems from this definition:
```c++
// gsl/gsl-lite.hpp, l. 838ff
#if gsl_HAVE( NORETURN )
# define gsl_NORETURN  [[noreturn]]
#elif defined(_MSC_VER)
# define gsl_NORETURN  __declspec(noreturn)
#elif gsl_COMPILER_GNUC_VERSION || gsl_COMPILER_CLANG_VERSION || gsl_COMPILER_APPLECLANG_VERSION || gsl_COMPILER_ARMCC_VERSION
# define gsl_NORETURN  __attribute__((noreturn))
#else
# define gsl_NORETURN
#endif
```
If NVHPC can deal with libstdc++ I assume it will handle most of GCC's `__attribute__()` annotations, so adding `|| gsl_COMPILER_NVHPC_VERSION` to the list should fix that warning, too.

Including \<endian.h\> like this is probably fine on Linux as long as NVHPC is documented to use g++'s standard library. I assume, however, that it will not work for NVHPC on Windows (which is not yet released but in the works as far as I know); the `#if defined( _WIN32 )` branch in the definition of `enum class endian` will take care of that, but perhaps we could generally avoid including the header file on Windows, i.e.
```diff
- #if gsl_HAVE( ENUM_CLASS ) && ( gsl_COMPILER_ARMCC_VERSION || gsl_COMPILER_NVHPC_VERSION )
+ #if gsl_HAVE( ENUM_CLASS ) && ( gsl_COMPILER_ARMCC_VERSION || gsl_COMPILER_NVHPC_VERSION ) && !defined( _WIN32 )
  # include <endian.h>
  #endif
```

One more thing: `nvc++` has a `--⁠no_exceptions` switch which we should ideally be able to detect. Again, because the compiler seems to support libstdc++, it will probably behave like GCC in this regard, i.e. it will define `__cpp_exceptions` only if exceptions are enabled. The following patch should work:
```diff
  // gsl-lite.hpp, l. 553ff
  #elif gsl_COMPILER_GNUC_VERSION
  # if gsl_BETWEEN(gsl_COMPILER_GNUC_VERSION, 1, 500)
  #  ifdef __EXCEPTIONS
  #   define gsl_HAVE_EXCEPTIONS  1
  #  else
  #   define gsl_HAVE_EXCEPTIONS  0
  #  endif // __EXCEPTIONS
  # else
  #  ifdef __cpp_exceptions
  #   define gsl_HAVE_EXCEPTIONS  1
  #  else
  #   define gsl_HAVE_EXCEPTIONS  0
  #  endif // __cpp_exceptions
  # endif // gsl_BETWEEN(gsl_COMPILER_GNUC_VERSION, 1, 500)
+ #elif gsl_COMPILER_NVHPC_VERSION
+ # ifdef __cpp_exceptions
+ #  define gsl_HAVE_EXCEPTIONS  1
+ # else
+ #  define gsl_HAVE_EXCEPTIONS  0
+ # endif // __cpp_exceptions
  #elif gsl_COMPILER_MSVC_VERSION
```

Would you mind trying this out?

martinmoene:
With respect to the warning `"<unnamed>::__lest_registrar__694" was declared but never referenced`, the suggested change to *lest* does not compile:

```diff
  #define lest_CASE( specification, proposition ) \
      static void lest_FUNCTION( lest::env & ); \
-     namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); } \
+     namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); (void) lest_REGISTRAR; } \
      static void lest_FUNCTION( lest_MAYBE_UNUSED( lest::env & lest_env ) )
```

Perhaps something along the lines of the following is needed:

```diff
  #define lest_CASE( specification, proposition ) \
      static void lest_FUNCTION( lest::env & ); \
-     namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); } \
+     namespace { lest::add_test lest_REGISTRAR( specification, lest::test( proposition, lest_FUNCTION ) ); lest::use(lest_REGISTRAR); } \
      static void lest_FUNCTION( lest_MAYBE_UNUSED( lest::env & lest_env ) )
```

with `lest_use()` like:

```cpp
namespace lest {

  template< typename T >
  void use( T const & ) {}

}
```

Not entirely sure if such implementation suffices.

I'd be happy to add such a change to *lest*.

mbeutel:
Ouch. Indeed. Thanks for the correction.

mbeutel:
(Speaking of *lest*: please note that I made some ad-hoc changes to our local copy in 39ce5470a73f99e8174b563244a846270bd9382b, 2401636dff35aa158417535deae434eda13141b6, 29ec9fbabf9b0f29b485e1a7b5c53b31a6e5f66a (stupidly done in the `master` branch instead of a dedicated PR), so although I'm in favor of staying in sync with upstream *lest*, we should take care to preserve these changes.

The changes in question were made to address #264; as explained at https://github.com/gsl-lite/gsl-lite/issues/264#issuecomment-803660196, our copy of *lest* now supports *compilation* but not *execution* without exception support; the latter would have required fundamental changes because *lest*'s design is based on exceptions. Being just a stop-gap for a specific need of *gsl-lite*, these changes are probably inappropriate for merging into upstream *lest*.)

martinmoene:
Thanks :)

I'll look at it later.

Have to get *lest* over to travis.com from travis.org (seems the only? repo not found there, still not after syncing with GitHub).

test/assert.t.cpp
```diff
@@ -36,6 +36,9 @@ enum Color
 #endif
 { red, green, blue };
 
+#if gsl_COMPILER_NVHPC_VERSION
+# pragma diag_suppress 941 // Suppress:  missing return statement at end of non-void function "..."
```
mbeutel, line 40: The entire point of this test case is to prove that the code compiles warning-free if `gsl_FailFast()` is used, so I'd suppress it here only as a last resort. Does the `__attribute__((noreturn))` annotation in `fail_fast_terminate()` and `fail_fast_throw()` not suffice to suppress the warning?

olupton: Unfortunately it doesn't seem to. I also tried with an explicit NVHPC clause in
```
#elif gsl_COMPILER_GNUC_VERSION || gsl_COMPILER_CLANG_VERSION || gsl_COMPILER_APPLECLANG_VERSION || gsl_COMPILER_ARMCC_VERSION || gsl_COMPILER_NVHPC_VERSION
# define gsl_NORETURN  __attribute__((noreturn))
#else
```

I obviously agree the suppression isn't a nice solution, I just thought I'd push the result of testing your suggestions. The current head of the PR builds without warnings for me, and the tests pass.

mbeutel: Oh well. Thanks for checking.

mbeutel:
Another issue is that our CMake compiler detection logic in https://github.com/gsl-lite/gsl-lite/blob/master/test/MakeTestTarget.cmake isn't aware of the NVHPC compiler and hence doesn't enable C++11/14/17 support for it. I guess we should let CMake detect language version support and just [`CMAKE_CXX_COMPILE_FEATURES`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_COMPILE_FEATURES.html) anyway; I'll get that done in the next days.

Even if `__attribute__((noreturn))` doesn't suffice to suppress the warning, I'd hope the standard C++11 `[[noreturn]]` attribute does; in this case the warning suppression should be restricted to pre-C++11 mode. Let's revisit this after I updated the build script.

olupton:
I tried to squash a few more complaints with other C++ versions enabled.
With C++11 and later then `[[noreturn]]` seems to work fine (but using it in C++98 and suppressing the "this is a C++11 feature" warning does not...).

mbeutel:
Pardon me for not getting back to this sooner.

Thank you for the thorough PR. I'm not too concerned about the warning as long as it occurs only pre-C++11. I think that some follow-up changes may be in order (in particular, modernizing the CMake build script), but that's no reason to further delay this PR.

PR #312, opened 2021-09-29
Add `gsl::make_unique<>()`, `gsl::make_shared<>()` which return `not_null<...>`

mbeutel:
The result of a `make_shared<T>()` and `make_unique<T>()` can never be `nullptr`. So it makes sense for the *gsl-lite* version of these functions to return `not_null<shared_ptr<T>>` and `not_null<unique_ptr<T>>`.

This may be a breaking change for users of `gsl::make_unique<T>()`, which previously forwarded to `gsl::std14::make_unique<T>()`, which returned and still returns `unique_ptr<T>`. However, in most cases the change should not necessitate source code changes because `not_null<unique_ptr<T>>` implicitly converts to `unique_ptr<T>`.

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -1595,10 +1595,6 @@ typedef gsl_CONFIG_INDEX_TYPE diff;
 #if gsl_HAVE( SHARED_PTR )
   using std::unique_ptr;
   using std::shared_ptr;
-  using std::make_shared;
-# if gsl_HAVE( MAKE_UNIQUE ) || gsl_HAVE( VARIADIC_TEMPLATE )
-  using std14::make_unique;
-# endif
 #endif
 
 #if  gsl_HAVE( ALIAS_TEMPLATE )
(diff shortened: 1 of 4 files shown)


include/gsl/gsl-lite.hpp
```diff
@@ -3276,6 +3272,26 @@ std::basic_ostream< CharType, Traits > & operator<<( std::basic_ostream< CharTyp
 }
 
 
+#if gsl_HAVE( VARIADIC_TEMPLATE )
+# if gsl_HAVE( UNIQUE_PTR )
+template< class T, class... Args >
+gsl_NODISCARD not_null<std::unique_ptr<T>>
+make_unique( Args &&... args )
+{
+    return not_null<std::unique_ptr<T>>( new T( std::forward<Args>( args )... ) );
```
bugwelle, line 3281: What if the user has overloaded `new` for a specific class? What `operator new()` would be called?

See https://godbolt.org/z/57cb1rfEY

```cpp
class Test
{
public:
    Test(std::string str) {   }

    void * operator new(std::size_t size) noexcept
    {
        return nullptr;
    }
};
```

As far as I can see, it _could_ be possible that `new` returns `nullptr`. For example in cases where exceptions are not desired and `nullptr` is returned in case of out-of-memory cases (or something similar; no idea why someone would use it that way).

bugwelle: Also, if I were to write `auto t2 = std::make_unique<Test>("abc");`, the unique ptr would be a `nullptr` :-/

Example: https://godbolt.org/z/fhxhbKvYW

mbeutel: > What if the user has overloaded `new` for a specific class? What `operator new()` would be called?

The overloaded operator would be called. No surprises there; our `make_unique<>()` is exactly equivalent to [`std::make_unique<>()`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) in that regard. We just need to duplicate the implementation rather than calling the `std` version because `std::make_unique<>()` didn't exist before C++14.


> As far as I can see, it _could_ be possible that `new` returns `nullptr`.

Sure. But the `not_null<>` constructor checks its preconditions, so this won't go unnoticed. By default, the application would terminate.


Bottom line: if you use `make_unique<>()` with overloaded `operator new`, and your program is prepared for it to return `nullptr`, just use `std::make_unique<>()` or the `gsl::std14::make_unique<>()` polyfill. `gsl::make_unique<>()` is for users of `not_null<>`.

bugwelle: Thanks for your reply. Sure, I only added the example because of your statement:

> The result of a `make_shared<T>()` and `make_unique<T>()` can never be `nullptr`

Which isn't correct 100% of the time (just a small nitpick) . :)

But of course, for the `not_null<>` case it makes sense that an exception is thrown if it returns `nullptr`.

PR #315, opened 2021-10-13
Check for gsl:: target in Config.cmake

Ram-Z:
Currently this file will be generated into a target `gsl-lite::gsl-lite` which doesn't exist.

I don't think this check is even needed, AFAIU the `targets.cmake` file will do this check also.


mbeutel:
Thank you for the contribution.

You are right about the check being unnecessary. Generally, our CMake build scripts could use some modernization.

PR #321, opened 2022-02-08
add SYSTEM to target_include_directories

mrahn:
In order to let compilers skip warnings caused by
gsl-lite. (e.g. `clang`s `-Wweak-vtables` is triggered by `fail_fast`)

See https://github.com/mrahn/test_gsl_lite_weak_vtables for a usage that triggers `-Wweak-vtables`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ add_library(
     ${package_nspace}::${package_name} ALIAS ${package_name} )
 target_include_directories(
     ${package_name}
+    SYSTEM
     INTERFACE
         "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
         "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>" )
@@ -89,6 +90,7 @@ add_library(
     ${package_nspace}::${package_name}-v0 ALIAS ${package_name}-v0 )
 target_include_directories(
     ${package_name}-v0
+    SYSTEM
     INTERFACE
         "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
         "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>" )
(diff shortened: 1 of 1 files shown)


mrahn:
Hi and thanks for your reply.

I do agree with your reasoning and the suggestion to use `pragma`  to suppress the warning locally. And I am happy that you share the goal to have code without warnings. And by the way thanks a lot for your contributions, they help a lot in producing modern code in restricted environments.

For my current work the, lets call it, work-around in the PR is sufficient and maintainable and I sadly can not promise to spend more time to come up with a proper upstream fix that takes into account the aspects you mention.

So I will close the PR. Not sure the topic is worth it to open an issue!?

Again, thanks a lot for your work.

mbeutel:
Thank you for your understanding.

I'd appreciate if you could open an issue so this doesn't get lost. It is not a big change to make, and I'll resolve this when I take the time for another round of updates, probably by the end of the month.

mrahn:
Okay, I opened https://github.com/gsl-lite/gsl-lite/issues/322 to avoid the `}`. Thanks!

wirew0rm:
Hey, thanks for already thinking about this and explaining your reasoning why you opted to not go into this direction.

I also wanted to request this feature as we are consuming gsl-lite using cmakeFetchContent. Because it is not a direct dependency, it is also not really feasable to modify the warnings where the headers are used.

Imho it would be acceptable for people that opt to set up their include path manually to also have to handle warnings for themselves.
Regarding fixing warnings instead of turning them off, I think that is a noble goal, which we also aim for, but as a mere user you usually cannot quickly fix such warnings and thus it prevents us from enforcing our own codebase to be warning free with -Werror. But of course, the testcases and examples shipped with gsl-lite should be build against a target with all reasonable warnings enabled (this seems to be missing in this PR). That way you can fix all warnings but do not block your users from enforcing stricter/newer warnings.

Maybe a good solution would be to have a flag that allows users to enable the `SYSTEM`flag. A reference for a popular header-only library following this approach is fmt: https://github.com/fmtlib/fmt/blob/ed18ca3eaec6c6b3d3c281b861476d4f5a297e79/CMakeLists.txt#L84
If you think that would be a viable solution, I could provide a PR.

I'll also open a bug report about the warning in question, just in case that this is something that actually can be fixed in the library.

PR #326, opened 2022-05-03
Fix `not_null_ic` copy construction

runer112:
Without this change, copy construction caused infinite recursion.

Reproduction including a toggle between old and new behavior: https://godbolt.org/z/o9W19o9nh.

### DIFF
--- a/include/gsl/gsl-lite.hpp
+++ b/include/gsl/gsl-lite.hpp
@@ -2660,8 +2660,8 @@ public:
         , typename std::enable_if< ( std::is_constructible<T, U>::value ), int >::type = 0
 #  endif
     >
-    gsl_api gsl_constexpr14 explicit not_null( U other )
-    : data_( T( std::move( other ) ) )
+    gsl_api gsl_constexpr14 explicit not_null( U && other )
+    : data_( T( std::forward<U>( other ) ) )
     {
         gsl_Expects( data_.ptr_ != gsl_nullptr );
     }
@@ -2682,8 +2682,8 @@ public:
         // We *have* to use SFINAE with an NTTP arg here, otherwise the overload is ambiguous.
         , typename std::enable_if< ( std::is_constructible<T, U>::value && !std::is_convertible<U, T>::value ), int >::type = 0
     >
-    gsl_api gsl_constexpr14 explicit not_null( U other )
-    : data_( T( std::move( other ) ) )
+    gsl_api gsl_constexpr14 explicit not_null( U && other )
+    : data_( T( std::forward<U>( other ) ) )
     {
         gsl_Expects( data_.ptr_ != gsl_nullptr );
     }


mbeutel:
Thank you for the PR and for the repro.

While `not_null<>` has undergone significant changes over the last few releases, `not_null_ic<>` has remained largely untouched. We have no test coverage for it as far as I can tell.

As far as I can remember, my reason for wanting to avoid constructors with forwarding references was exactly the interference with copy constructors which is at the root of the problem you observed:
https://mpark.github.io/programming/2014/06/07/beware-of-perfect-forwarding-constructors/

Therefore, I think a better remedy would be to leave `not_null<>` intact and change `not_null_ic<>` instead:

```diff
template< class T >
class not_null_ic : public not_null<T>
{
public:
    template< class U
        gsl_ENABLE_IF_(( std::is_constructible<T, U>::value ))
    >
    gsl_api gsl_constexpr14
#if gsl_HAVE( MOVE_FORWARD )
-    not_null_ic( U && u )
-    : not_null<T>( std::forward<U>( u ) )
+    not_null_ic( U u )
+    : not_null<T>( std::move( u ) )
#else // ! gsl_HAVE( MOVE_FORWARD )
    not_null_ic( U const & u )
    : not_null<T>( u )
#endif // gsl_HAVE( MOVE_FORWARD )
    {}
};
```

runer112:
Yeah, it looks like that solution works, too. But I must say that I'm not too keen on the unnecessary moves (or if the type is not moveable for some reason, **copies**) introduced by only accepting values. A primary goal of the `not_null` wrapper is to be as lightweight as possible, correct?

The post you linked to suggests SFINAE as a solution, which it looks like is already applied to the type-generic `not_null` constructors (with extra logic for differentiating constructible vs. convertible). Does this not sufficiently address the issue? I see that the SFINAE section has a footnote about ["perfect forwarding failure cases"](https://groups.google.com/forum/#!topic/comp.std.c++/wu4r3CWG-VE), but I don't think any of the cases illustrated apply here?

mbeutel:
I suppose you're right. Let's do it this way, then.

mbeutel:
One thing that holds me back is that I don't understand why we end up with a recursive call in the current version. In your repro, MSVC emits the following warning:

```
<source>(3289) : warning C4717: 'gsl::not_null_ic<std::shared_ptr<int> >::not_null_ic<std::shared_ptr<int> ><gsl::not_null_ic<std::shared_ptr<int> > &,0>': recursive on all control paths, function will cause runtime stack overflow
```
for the following (de-cluttered) code:
```c++
    template< class U
        gsl_ENABLE_IF_(( std::is_constructible<T, U>::value ))
    >
    gsl_api gsl_constexpr14
    not_null_ic( U && u )
    : not_null<T>( std::forward<U>( u ) )
    {}  // <-- line 3289
```
We have `T = std::shared_ptr<int>` and `U = gsl::not_null_ic<std::shared_ptr<int>> &`. `std::forward<T>(x)` is essentially `static_cast<T&&>(x)`, so above code is read as
```c++
    gsl_api gsl_constexpr14
    not_null_ic( gsl::not_null_ic<std::shared_ptr<int>> & u )
    : not_null<std::shared_ptr<int>>( static_cast<gsl::not_null_ic<std::shared_ptr<int>> &>( u ) )
    {}
```
How does the recursive call happen?

And another thing: `std::is_constructible<std::shared_ptr<int>, not_null_ic<std::shared_ptr<int>>&>::value` evidently evaluates to `true`, hence the constructor still takes precedence over the copy constructor despite our SFINAE constraint. So I think the constraint needs to be refined. I suppose we also need a set of `not_null_ic<T>(not_null<U>)` overloads then.

It should not be surprising that the SFINAE constraint is not sufficient. According to the documentation, the [`is_constructible<>`](https://en.cppreference.com/w/cpp/types/is_constructible) and [`is_convertible<>`](https://en.cppreference.com/w/cpp/types/is_convertible) traits have undefined behavior when used with incomplete types. Varying behavior can be observed in practice:
https://gcc.godbolt.org/z/d7qnnqo9q
Observe how the results differ when commenting/uncommenting lines 20/21.

---

Given all this complexity, perhaps it would be preferable to use a by-value argument?

runer112:
Hmm... Yeah, I guess this does get rather complicated. For the sake of simplicity, I guess I'd be okay with your initial counter-proposal to change `not_null_ic` to have an lvalue-only constructor that moves.

runer112:
Updated playground for testing alternate/original/no fix: https://godbolt.org/z/cPfoeM6cv

mbeutel:
Thanks.

Before I merge this, would you mind adding your test case to our test suite so we can make sure we don't regress?

mbeutel:
Pardon me for forgetting about this.

For the test case we can take advantage of the fact that MSVC detects the recursion at compile time and issues a warning. Otherwise, infinite recursion quickly overflows the stack at runtime, leading to a failed test.

I'll merge this PR now and add some test coverage for `not_null_ic<>` later.

PR #328, opened 2022-06-01
Repair the CI pipelines

mbeutel:
They stop working if we don't groom them from time to time.


mbeutel:
The AVX2-related false positives in memcheck remain, not sure what to do about these. Github gives us occasional "ServerBusyEgress" errors when downloading Ninja; we could cache it if the problem persists, but that would be something to address in CImakeshift. So I'll merge this PR now despite the failures.

PR #338, opened 2023-04-18
Cast char to signed to avoid type-limits warning

### LINKED ISSUE #337: Tests fail to compile due to -Werror=type-limits when char is unsigned
On Fedora Linux Rawhide, with GCC 13.0.1, while testing gsl-lite 0.41.0, we’re seeing that `test/lest_cpp03.hpp` fails to compile with an error from `-Werror=type-limits` on our non-x86 architectures (`aarch64`, `ppc64le`, `s390x`).

```
In file included from /builddir/build/BUILD/gsl-lite-0.41.0/test/gsl-lite.t.hpp:104,
                 from /builddir/build/BUILD/gsl-lite-0.41.0/test/emulation.t.cpp:20:
/builddir/build/BUILD/gsl-lite-0.41.0/test/lest_cpp03.hpp: In function 'bool lest::unprintable(char)':
/builddir/build/BUILD/gsl-lite-0.41.0/test/lest_cpp03.hpp:707:46: error: comparison is always true due to limited range of data type [-Werror=type-limits]
  707 | inline bool unprintable( char c ) { return 0 <= c && c < ' '; }
      |                                            ~~^~~~
```

This seems to be happening because `char` is equivalent to `signed char` on `i686` and `x86_64`, but is equivalent to `unsigned char` on the other platforms. The useless comparison is not a real problem here, but the failure to compile is.

I am not sure why we are seeing this in 0.41.0 but not in 0.40.0. Nothing stands out in the [changes](https://github.com/gsl-lite/gsl-lite/compare/v0.40.0...v0.41.0) as potentially relevant.

There is existing logic to suppress this warning on “`GNU`” compilers, but only for old compiler versions:

https://github.com/gsl-lite/gsl-lite/blob/755ba124b54914e672737acace6a9314f59e8d6f/test/MakeTestTarget.cmake#L264-L266

Should we just make that unconditional, like `-Wno-long-long` and `-Wuseless-cast`?

musicinmybrain:
On at least GCC 13, on platforms where `char` is unsigned by default, checking if a `char`’s value is nonnegative by comparing to integer zero produced a warning which was treated as an error: “`comparison is always true due to limited range of data type [-Werror=type-limits]`”

Fixes #337.

### DIFF
--- a/test/lest_cpp03.hpp
+++ b/test/lest_cpp03.hpp
@@ -704,7 +704,7 @@ inline void inform( location where, text expr )
 
 // Expression decomposition:
 
-inline bool unprintable( char c ) { return 0 <= c && c < ' '; }
+inline bool unprintable( char c ) { return 0 <= static_cast<signed char>(c) && c < ' '; }
 
 inline std::string to_hex_string(char c)
 {
--- a/test/lest_cpp03.hpp
+++ b/test/lest_cpp03.hpp
@@ -704,7 +704,7 @@ inline void inform( location where, text expr )
 
 // Expression decomposition:
 
-inline bool unprintable( char c ) { return 0 <= static_cast<signed char>(c) && c < ' '; }
+inline bool unprintable( char c ) { return 0 <= static_cast<signed char>(c) && c < ' '; }  // the cast is necessary on architectures where char is unsigned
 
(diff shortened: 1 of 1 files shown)


mbeutel:
Thank you. I added a comment and will merge the PR.

PR #354, opened 2025-05-13
v1.0.0

### LINKED ISSUE #153: Static extent in span
As far as I know there is still no proper definition of the GSL (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#faq51-is-githubcommicrosoftgsl-the-gsl). But it seems to me that it makes sense for the definition of span to follow what's being standarised for C++20.

gsl-lite::span doesn't support static extents (https://github.com/martinmoene/gsl-lite/blob/master/include/gsl/gsl-lite.hpp#L451), which is more strange given that:

- The Microsoft GSL implementation does support static extents (https://github.com/Microsoft/GSL/blob/master/include/gsl/span#L75)
- The gsl-lite limitation doesn't seem to be because of a lack of resources since span-lite does support static extents (https://github.com/martinmoene/span-lite/blob/master/include/nonstd/span.hpp#L496)

### LINKED ISSUE #179: Binary compatibility
*gsl-lite* currently defines several switches and types in ways which can lead to breakage of binary compatibility if different configuration settings or different language standard settings are used in different translation units:

- The `byte` type, which is either an `enum class` (C++11) or a `struct` (C++98).
- `gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF` alters the return type of a function.
- `final_action`, `final_action_return`, and `final_action_error` are either class templates (C++11) or classes (C++98). This also affects the return values of `finally()`, `on_return()`, and `on_error()`.
- `std11::integral_constant<>`, which is either an alias for `std::integral_constant<>` (pre-C++11) or a dedicated type.
- The `gsl_api` macro (e.g. `-Dgsl_api="__stdcall"` can lead to link-time errors).
- `gsl_CONFIG_SPAN_INDEX_TYPE` (mixing different configuration settings can lead to link-time errors)

Comment https://github.com/martinmoene/gsl-lite/issues/121#issuecomment-524560523 illustrates how different build settings in separate targets that are linked together can lead to link-time failure (or even run-time data corruption in contrived scenarios).

Proposed resolutions:

- Don't define `gsl::byte` at all for C++98. Using `struct` would lead to suboptimal codegen anyway because `struct` arguments aren't commonly passed in registers, so perhaps it just isn't worth the trouble. Or if a byte type for C++98 is strongly desired, we should name it `struct byte98` and define it unconditionally.
- Make `gsl_CONFIG_NOT_NULL_GET_BY_CONST_REF` default and non-configurable. (I see no downsides to this; please let me know if I'm missing something.)
- Instead of non-templated `final_action`, define `final_action<>` with a default template parameter `T = void (*)()`, so C++98 code can say `final_action<> a = finally(cleanup_func);`.
- The current definition of `std11::integral_constant<>` is not a problem if we only use it for metaprogramming (as we currently do, e.g. in `is_basic_string_span<>`), but then we should put it in the `detail` namespace so users don't depend on it. If we want it to be public, it should always be a dedicated type unrelated to `std::integral_constant<>`.
- If the programmer customizes `gsl_api` or `gsl_CONFIG_SPAN_INDEX_TYPE`, we should emit a warning that says "You are using gsl-lite with configuration options that affect the binary interface. You probably shouldn't do this if you are building a library because this can cause link-time errors for users of your library who also want to use gsl-lite. Make sure you control the build flags of all components linked into your target.". The warning can then be suppressed by defining something like `gsl_ACKNOWLEDGE_NONSTANDARD_ABI`.

### LINKED ISSUE #194: Collisions with M-GSL

### LINKED ISSUE #198: Add documentation
The Guidelines Support Library is sort-of documented in the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-gsl), so one could argue that GSL implementations such as *gsl-lite* only need to document their configuration switches.

However, *gsl-lite* offers a variety of additional features which are currently undocumented (e.g. the `gsl_HAVE_*` family of macros, language level-sensitive macros like `gsl_nullptr`, or `ExpectsAudit()`).

I think we should add dedicated documentation for all of *gsl-lite*'s features and some guidelines for proper use (cf. #194). 

My current plan is to use Sphinx for this, perhaps together with Doxygen and Breathe.

### LINKED ISSUE #245: Ambiguous references to `gsl` in CUDA device code
```cpp
// example.cu
#include <gsl-lite/gsl-lite.hpp>

namespace foo {

namespace gsl = ::gsl_lite;

__global__ void kernel(gsl::span<float> s) {}

void func(gsl::span<float> s) {}

}

using namespace foo;

int main() {
}
```

The compilation with `nvcc` 10.2.89 fails with:

```
In file included from tmpxft_0000a58f_00000000-5_gslnamespace.cudafe1.stub.c:1:
/tmp/tmpxft_0000a58f_00000000-5_gslnamespace.cudafe1.stub.c:10:9: error: reference to ‘gsl’ is ambiguous
 typedef gsl::span<float>  _ZN3gsl4spanIfEE;
```

Please note that the definition of function `func` works totally fine. I do not think that this is an error caused by the implementation of the gsl-lite, but rather a compilation issue of `nvcc`. Maybe it should be noted that you should not define the convenient namespace alias `namespace gsl = ::gsl_lite` for CUDA kernel functions (like it is described [here](https://github.com/gsl-lite/gsl-lite#using-gsl-lite-in-libraries)).

### LINKED ISSUE #282: basic_zstring_span::empty() returns false for an empty string
https://github.com/gsl-lite/gsl-lite/blob/503b14bdd1cfa9a797dc3dddab6baac7e13504bf/include/gsl/gsl-lite.hpp#L4309

Since basic_zstring_span operates on a null terminated strings then the basic_zstring_span::empty() should return true for 
`span_.size() <= 1;` since the span of `""` has the size of 1.

You have this logic here:
https://github.com/gsl-lite/gsl-lite/blob/503b14bdd1cfa9a797dc3dddab6baac7e13504bf/include/gsl/gsl-lite.hpp#L4317
https://github.com/gsl-lite/gsl-lite/blob/503b14bdd1cfa9a797dc3dddab6baac7e13504bf/include/gsl/gsl-lite.hpp#L4324
https://github.com/gsl-lite/gsl-lite/blob/503b14bdd1cfa9a797dc3dddab6baac7e13504bf/include/gsl/gsl-lite.hpp#L4295

### LINKED ISSUE #309: Support `narrow<>()` for types without total order
Cf. https://github.com/microsoft/GSL/pull/986.

mbeutel:
Tentative release notes: https://github.com/gsl-lite/gsl-lite/releases/tag/v1.0.0-prerelease-1

This PR also adds documentation, which is already live at https://gsl-lite.github.io/gsl-lite/.

Closes #245, closes #194 (migrated to namespace `gsl-lite`).
Closes #153 (added static extent to `span<>`).
Closes #198 (added documentation).
Closes #179 (implemented safety measures against unintentional ABI breaks).
Closes #309 (`narrow<>()` was implemented for types without total order).
Closes #282 (`string_span` was deprecated before and is now opt-in).

@martinmoene, @petamas, @pauleonix, @mpusz, @BenFrantzDale, and everyone else interested: Would you mind taking a look at the [feature/v1.0](https://github.com/gsl-lite/gsl-lite/tree/feature/v1.0) branch and the [new documentation](https://gsl-lite.github.io/gsl-lite/) and tell me what you think? I'd be super grateful if some of you could give the new version a test run before I release it.


martinmoene:
Wow, congratulations on reaching this 1.0 milestone, all the work, stepwisely refining, leading up to it.

Could not have imagined how well this (handing over the project) turns out, also becoming a nice example to learn things from :)

I may not have anything substantial to add. Scan-read the nice new documentation and successfully built the tests and examples locally with CMake / Visual Studio 16 2019.

Thank you so much!

mbeutel:
Thank you for your kind words, Martin.

I'm merging this now and hope to finish the v1.0.0 release later today.

PR #359, opened 2025-09-04
add fix for nvhpc compiler

wuyifff:
## Description
This PR fixes build failures with the NVHPC compiler. Like NVCC, it does not handle some advanced C++ template features.

## Changes

1. Disabled a not_null accessor method for the NVHPC compiler to prevent a compilation error.
2. Applied an existing workaround for NVCC not handling SFINAE to also cover the NVHPC compiler, which has the same issue.

### DIFF
--- a/include/gsl-lite/gsl-lite.hpp
+++ b/include/gsl-lite/gsl-lite.hpp
@@ -2790,13 +2790,13 @@ struct not_null_elem
 {
     typedef typename element_type_helper<T>::type element_type;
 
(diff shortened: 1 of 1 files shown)


mbeutel:
> One question, though. I understand that NVHPC has the same SFINAE issue as NVCC with regard to `not_null<>::operator()`. But which problem do you experience with `not_null<>::get()`, which does not rely on SFINAE? I'm wondering if we could introduce a workaround for `not_null<>::get()` so we don't have to disable it entirely for NVHPC.

@wuyifff Could you comment on this?

